PR #6, opened 2017-08-01
Navigation help profile

Mahmud-Khan:
### Summary
Added components for a Help and Settings menu of the navigation component.
desktop

mobile


packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
 const cx = classNames.bind(styles);
 
 const propTypes = {
   test: PropTypes.string,
-  help: PropTypes.string,
+  help: PropTypes.arrayOf(PropTypes.object),
   isMobileNavOpen: PropTypes.bool, // this is proof of concept. should we keep this?
   quickLinks: PropTypes.arrayOf(PropTypes.object),
   navItems: PropTypes.arrayOf(PropTypes.object),
+  profile: PropTypes.object,
```
mhemesath: Please use `Proptypes.shape` to validate the proper object structure is passed https://facebook.github.io/react/docs/typechecking-with-proptypes.html

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
+import UserIcon from './icons/UserIcon';
+import OutlineEllipses from './icons/OutlineEllipses';
 import styles from './Nav.scss';
 
 const cx = classNames.bind(styles);
 
 const propTypes = {
   test: PropTypes.string,
-  help: PropTypes.string,
+  help: PropTypes.arrayOf(PropTypes.object),
```
mhemesath: please verify object share  https://facebook.github.io/react/docs/typechecking-with-proptypes.html

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
   isMobileNavOpen: PropTypes.bool, // this is proof of concept. should we keep this?
   quickLinks: PropTypes.arrayOf(PropTypes.object),
   navItems: PropTypes.arrayOf(PropTypes.object),
+  profile: PropTypes.object,
   mainContent: PropTypes.node,
 };
 
 const defaultProps = {
   isMobileNavOpen: false,
   test: 'hello world',
```
mhemesath: remove this

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
+  help: PropTypes.arrayOf(PropTypes.object),
   isMobileNavOpen: PropTypes.bool, // this is proof of concept. should we keep this?
   quickLinks: PropTypes.arrayOf(PropTypes.object),
   navItems: PropTypes.arrayOf(PropTypes.object),
+  profile: PropTypes.object,
   mainContent: PropTypes.node,
 };
 
 const defaultProps = {
   isMobileNavOpen: false,
```
mhemesath: why is this needed?

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
+import NavHelp from './components/nav-help/NavHelp';
+import UserProfile from './components/user-profile/UserProfile';
+import OutlineInfo from './icons/OutlineInfo';
+import UserIcon from './icons/UserIcon';
+import OutlineEllipses from './icons/OutlineEllipses';
 import styles from './Nav.scss';
 
 const cx = classNames.bind(styles);
 
 const propTypes = {
```
mhemesath: Is this actually the layout, and not the nav? This component really needs to be renamed. Also, default proptypes shoudl be set in the examples... not in the actual component

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -3,23 +3,81 @@ import PropTypes from 'prop-types';
...
   navItems: PropTypes.arrayOf(PropTypes.object),
+  profile: PropTypes.object,
   mainContent: PropTypes.node,
 };
 
 const defaultProps = {
   isMobileNavOpen: false,
   test: 'hello world',
-  help: 'help',
+  help: [{
```
mhemesath: move all of this into an example page

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -0,0 +1,85 @@
+@import '../../variables';
```
mhemesath: fix the indentation in this file please

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -0,0 +1,85 @@
...
+	
+	.profile {
+	 font-size: 16px;
+	 line-height: 20px;
+	 padding-top: 20px;
+	 padding-bottom: 20px;
+	 padding-left: 25px;
+	 padding-right: 25px;
+	}
+	.profile-icon {
```
mhemesath: no need to prefix all this stuff with `profile`.... its localized to this file.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -0,0 +1,85 @@
...
+	  height: 16px;
+    width: 16px;
+    font-szie: 16px
```
mhemesath: you don't need height/width if font-size is spelled correctly :)

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,166 @@
...
+        onRequestClose={this.closePopup}
+        targetRef={() => document.getElementById('profile-link-icon')}
+        isArrowDisplayed
+        contentWidth="320"
+        contentHeight="400"
+        contentAttachment="top right"
+      >
+        {<div>
+          {profileLinks}
+          <div style={{ padding: '20px' }}>
```
mhemesath: no inline styles please

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,166 @@
...
+            <a className={cx('profile-link')} href={profile.signoutUrl} onClick={() => window.location = profile.signoutUrl} style={{ padding: '0px' }}>
+              {translations.help}
+            </a>
+          </div>
+          <div className={cx('divider-top')} style={{ padding: '20px' }}>
```
mhemesath: no inline stles

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,166 @@
...
+        targetRef={() => document.getElementById('profile-link-icon')}
+        isArrowDisplayed
+        contentWidth="320"
+        contentHeight="400"
+        contentAttachment="top right"
+      >
+        {<div>
+          {profileLinks}
+          <div style={{ padding: '20px' }}>
+            <a className={cx('profile-link')} href={profile.signoutUrl} onClick={() => window.location = profile.signoutUrl} style={{ padding: '0px' }}>
```
mhemesath: remove inline styles

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,166 @@
...
+              {translations.help}
+            </a>
+          </div>
+          <div className={cx('divider-top')} style={{ padding: '20px' }}>
+            <a className={cx('profile-link')} href={profile.signoutUrl} onClick={() => window.location = profile.signoutUrl} style={{ padding: '0px' }}>
```
mhemesath: remove inline styles

mhemesath:
Please remove all inline styles from this review

mhemesath:
For icons use [IconBase](https://github.com/cerner/terra-core/blob/master/packages/terra-icon/src/IconBase.jsx) from terra.

Here is an example of an icon built with it https://github.com/cerner/terra-core/blob/master/packages/terra-icon/src/icon/IconAdd.jsx

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: This is a container component that contains the navigation that includes profile, quick links & so on and the container component that contains the page content. Agree, Layout would be a more relevant name for this component.

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: Will pass the navigation as a prop. Should I pass the help navigations also as a prop explicitly? For the mobile layout, the help button is not visible and accessed from the Settings/Profile menu, so need the prop to call the Help component. It can be part of the array in navigation prop but sending it explicitly would be better I think.

mhemesath:
Please resolve conflicts in branch

packages/terra-consumer-nav/src/Layout.jsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
```
zhongyn: +1, layout is built by the consumer app based on nav components from this project

packages/terra-consumer-nav/src/Layout.scss
```diff
@@ -24,9 +24,22 @@
...
       max-width: 100vw;
       width: 100vw;
     }
   }
+  .footer {
+    width: 100%;
+    bottom: 15px;
+    margin:0;
+    padding:0;
+    position:sticky;
```
zhongyn: should it be fixed?

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -17,6 +16,44 @@ const propTypes = {
...
+    children: PropTypes.arrayOf(PropTypes.shape({
+      slug: PropTypes.string,
+      nav_type: PropTypes.oneOf(['GROUPING', 'EXTERNAL', 'APPLICATION', 'MODAL']),
+      text: PropTypes.string.isRequired,
+      uri: PropTypes.string.isRequired,
+      icon: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.element,
+      ]),
+      children: PropTypes.array,
```
zhongyn: is there any third level nav item?

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -64,4 +64,14 @@
...
   }
+
+  .profile-container {
+    bottom: 0;
+    height: auto;
+    left: 0;
+    position: absolute;
+    text-align: left;
+    width: inherit;
+    z-index: 200;
```
zhongyn: why is it 200? the general rule is ordering the z-index based on current stack context or local context: https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,121 @@
...
+import NavHelpContent from './NavHelpContent';
+import IconInfo from '../../icons/IconInfo';
+import styles from './NavHelp.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  help: PropTypes.arrayOf(PropTypes.shape({
+    slug: PropTypes.string,
+    nav_type: PropTypes.oneOf(['GROUPING', 'EXTERNAL', 'APPLICATION', 'MODAL']),
```
zhongyn: camelCase is preferred: navType

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  openPopup() {
+    this.setState({ isOpen: true });
+  }
+
+  closePopup() {
+    this.setState({ isOpen: false });
+  }
+
```
zhongyn: since the popup is either open or close, we can use one callback handling the state. e.g. https://github.cerner.com/HealtheLife/r_wellness/blob/master/client/examples/components/modal/DefaultModal.jsx#L12

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: There was a discussion and Mike told to ignore third level nav items. Please check this draft object nav is expecting and help item has children of children.
https://jira2.cerner.com/browse/PORTALDEV-34187

packages/terra-consumer-nav/src/Layout.scss
Mahmud-Khan: Based on UX callout the help button is positioned in the bottom left corner when the page content takes full screen but it would be moved up if the content does not fill up the full screen. Here is the desktop UI callout.
https://wiki.ucern.com/pages/viewpage.action?pageId=1332770962&preview=/1332770962/1623759251/healf_ui-uplift_navigation_callouts_v2.pdf

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,121 @@
...
+  render() {
+    const { help, ...customProps } = this.props;
+    const helpButton = (
+      <button
+        id="nav-help-button"
```
zhongyn: it should be a unique id passed by user instead of hardcoded. The popup component below will need to access this unique id. So if there are two help button. The popup will be confused

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
KevinJJackson: you can ignore slug, navtype and target. These are not used in our components

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,129 @@
+@import '../../variables';
+
+:local {
+	.nav-help {
```
zhongyn: indentation?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,129 @@
...
+    background-color:  #fff;
+    font-size: 16px;
+    line-height: 20px;
+    border-radius: 20px;
+    box-shadow: 0 1px 1px #666;
+    padding-top: 6px;
+    padding-bottom: 6px;
+    padding-left: 13px;
+    padding-right: 13px;
```
zhongyn: shorthand property:
```
padding: 6px 13px;
```

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,100 @@
...
+            fitStart={<div className={cx('icon')}>{content.icon}</div>}
+            fill={<div className={cx('padding-left-small')}>{content.text}</div>}
+            fitEnd={<div className={cx('icon')}>{toggleIcon}</div>}
+          />
+        </a>
+        <Toggler isOpen={!this.state.isOpen} isAnimated className={cx('padding-none')}>
+          { content.children.map((element) => {
+            const helpInfo = (<div key={`${element.text}`} className={cx('help-subitem', 'padding-top-md')}>{element.text}</div>);
+            return helpInfo;
+          })
```
zhongyn: it can be inline:
```
content.children.map((element) => (
  <div key={`${element.text}`} className={cx('help-subitem', 'padding-top-md')}>{element.text}</div>)
)
```

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,121 @@
...
+    })),
+  })),
+};
+
+const defaultProps = {
+  help: [],
+};
+
+const translations = {
+  help: 'Help',
```
KevinJJackson: is there a strategy for translations? terra-i18n?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,129 @@
...
+    font-size: 20px;
+    line-height: 30px;
+    color: #fff;
+    text-align: left;
+    padding-left: 15px;
+  }
+
+  .close-icon {
+    height: 16px;
+    width: 16px;
```
KevinJJackson: shouldn't need height/width when you have the font-size set

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,129 @@
+@import '../../variables';
+
```
KevinJJackson: make sure to run the linters, i'm seeing quite a few here that will throw red-flags

packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx
```diff
@@ -0,0 +1,45 @@
...
+const defaultProps = {
+  linkItems: [],
+};
+
+const ProfileLinks = ({
+  linkItems,
+  ...customProps
+}) => {
+  const linkElements = linkItems.map((linkItem) => {
+    const linkElement = (<a key={`${linkItem.text}`} href={linkItem.uri} onClick={() => window.location = linkItem.uri} className={cx('suppress-hyperlink')}>
```
KevinJJackson: do you have to set the onClick? isn't that what the href is for?

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,192 @@
...
+const propTypes = {
+  name: PropTypes.string.isRequired,
+  avatar: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  signoutUrl: PropTypes.string.isRequired,
+  profileLinks: PropTypes.arrayOf(PropTypes.shape({
+    slug: PropTypes.string,
+    nav_type: PropTypes.oneOf(['GROUPING', 'EXTERNAL', 'APPLICATION', 'MODAL']),
```
KevinJJackson: can remove slug and nav_type

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,192 @@
...
+    target: PropTypes.string,
+    text: PropTypes.string.isRequired,
+    uri: PropTypes.string.isRequired,
+    icon: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.element,
+    ]),
+    children: PropTypes.arrayOf(PropTypes.shape({
+      slug: PropTypes.string,
```
KevinJJackson: you are using this hash shape a lot. split it off into one constant and set the shape to it:
`children: PropTypes.arrayOf(PropTypes.shape(profileHashShape)),`

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,192 @@
...
+  constructor() {
+    super();
+
+    this.state = {
+      isOpen: false,
+      showModal: false,
+    };
+    this.showModal = false;
+    this.openProfilePopup = this.openProfilePopup.bind(this);
+    this.closeProfilePopup = this.closeProfilePopup.bind(this);
```
KevinJJackson: combine these into a toggle

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -22,7 +22,8 @@
       background: var(--terra-consumer--mobile-nav-background, #383f42);
       box-shadow: 0 1px 3px;
       display: block;
-      opacity: 0.95;
+      opacity: 0.8;
+      width: 100vw;
```
KevinJJackson: tablet is not supposed to have 100% screen width for the nav. remove these changes and i'll add the correct breakpoint changes with the other nav ui fixes.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,108 @@
...
+      contentWidth="320"
+      contentHeight="240"
+      classNameContent={cx('popup-content')}
+      contentAttachment="middle center"
+      popupContent={popupContent}
+    />);
+
+    return (
+      <div {...customProps}>
+        <ResponsiveElement responsiveTo="window" defaultElement={defaultElement} medium={popup} large={popup} />
```
KevinJJackson: you don't need to set the same content for medium and large, just set medium.

> It is not necessary to set each breakpoint. If a breakpoint is not set the component will look at each successive smaller breakpoint until one is found.
**http://terra-ui.herokuapp.com/components/core/responsive-element**

KevinJJackson:
css properties for background/font colors should be themable as a start and more will be available later (this is the same strategy used in the base Navigation component and the same style and naming convention for vars should be followed)

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,108 @@
...
+    const popupContent = <NavHelpContent helpContent={help} />;
+
+    const defaultElement = (<Modal
+      isModalOpen={this.state.isOpen}
+      title={translations.help}
+      content={popupContent}
+      closeModal={this.togglePopup}
+    />);
+
+    const popup = (<Popup
```
KevinJJackson: there is a weird tab index on this popup that needs to be removed. ex: (in chrome)
screen shot 2017-08-14 at 9 36 51 am

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -3,10 +3,12 @@
 
 :local {
   .nav-container {
+    background: inherit;
```
KevinJJackson: i don't think this is having the effect you are looking for completely, profile shouldn't stand out with its own background
screen shot 2017-08-14 at 9 48 38 am
:

Mahmud-Khan: I have noticed that also. This happens if the background is a gradient. Tested with a background color and in that case, profile gets same as the background. The mobile nav background is one color, in that case, it is working as expected. The only way I could get the profile to have the same color as the gradient background was making it transparent. But in that case, it causes an issue when there are more nav elements and scroll is present. I am trying to find a solution. Please let me know if you have a solution in mind that I can try. Thank you.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Need to discuss the i18n strategy. Working to use the terra-i18n component to wrap the layout component so that all children can get locale specific translation messages.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Removed slug and nav_type props as they are not used for Nav and Layout components.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Updated to use one callback for toggling popup.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Updated to pass the help button id as a prop of the Nav component.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
Mahmud-Khan: Fixed the indentation.

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
Mahmud-Khan: Made it inline.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
Mahmud-Khan: Fixed all lint issues.

packages/terra-consumer-nav/src/Nav.scss
Mahmud-Khan: Reviewed the Nav with Scott on Friday and he wanted the opacity changed to 0.8. I will remove this Nav UI changes.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
Mahmud-Khan: Updated

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
Mahmud-Khan: :) . Updated the font-size.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Removed all the inline styles. There is one to add sample content in the main container of the  Layout.jsx. This is used to test the help button position for the different case.The layout is now part of the test folder, so kept it for the test, can remove it if suggested.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Removed all the inline styles. There is one to add sample content in the main container of the Layout.jsx. This is used to test the help button position for the different case.The layout is now part of the test folder, so kept it for the test, can remove it if suggested.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Removed all the inline styles. There is one to add sample content in the main container of the Layout.jsx. This is used to test the help button position for the different case.The layout is now part of the test folder, so kept it for the test, can remove it if suggested.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Removed all the inline styles. There is one to add sample content in the main container of the Layout.jsx. This is used to test the help button position for the different case.The layout is now part of the test folder, so kept it for the test, can remove it if suggested.

packages/terra-consumer-nav/src/Layout.scss
Mahmud-Khan: Based on UX review, the help button does not require a sticky position. The help button has fixed position for desktop and the position is relative to page content in mobile/tablet view.

packages/terra-consumer-nav/src/Nav.scss
Mahmud-Khan: Removed the z-index.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
Mahmud-Khan: Will update it.

package.json
```diff
@@ -100,5 +100,8 @@
     "stylelint-scss": "1.5.1",
     "webpack": "^3.3.0",
     "webpack-dev-server": "^2.5.1"
+  },
+  "dependencies": {
+    "terra-grid": "^3.5.1"
```
alex-bezek: was this added on purpose? I don't think our root package.json should have dependencies should it?

packages/terra-consumer-nav/package.json
```diff
@@ -36,8 +36,14 @@
     "terra-button": "^1.1.0",
     "terra-icon": "^1.0.0",
     "terra-mixins": "^1.0.0",
+    "terra-modal": "^1.2.0",
```
alex-bezek: now that we are consuming `terra-modal` can we remove 
https://github.com/cerner/terra-consumer/tree/Navigation-help-profile/packages/terra-consumer-nav/src/components/modal

packages/terra-consumer-nav/package.json
```diff
@@ -36,8 +36,14 @@
...
+    "terra-overlay": "^1.1.0",
     "terra-status": "^1.1.0",
-    "terra-toggle": "^1.1.0"
+    "terra-toggle": "^1.1.0",
+    "terra-popup": "^1.0.0",
+    "terra-responsive-element":"^1.1.0",
+    "terra-grid":"^3.6.0",
+    "terra-icon":"^1.3.0"
```
alex-bezek: this is a duplicate

packages/terra-consumer-nav/package.json
```diff
@@ -36,8 +36,14 @@
...
     "terra-icon": "^1.0.0",
     "terra-mixins": "^1.0.0",
+    "terra-modal": "^1.2.0",
+    "terra-overlay": "^1.1.0",
     "terra-status": "^1.1.0",
-    "terra-toggle": "^1.1.0"
+    "terra-toggle": "^1.1.0",
+    "terra-popup": "^1.0.0",
+    "terra-responsive-element":"^1.1.0",
+    "terra-grid":"^3.6.0",
```
alex-bezek: np: can we keep these alphabetized ?

package.json
Mahmud-Khan: Added by mistake, will remove it.

packages/terra-consumer-nav/package.json
Mahmud-Khan: Did not use terra-modal, will remove it. UX callout requires the modal title to be on the overlay, so create this modal component using terra-overlay. It will be replaced when Terra-consumer has its own Modal.

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -62,6 +63,44 @@ const propTypes = {
     ),
   ),
   /**
+   * An array of items to be displayed as help menu content.
+   */
+  helpItems: PropTypes.arrayOf(PropTypes.shape(
```
alex-bezek: This seems spread out all over the components. Really you shouldn't have to declare the shape here if its just passing it along into sub components. Let them declare the shape and throw their prop type mismatch warnings

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -62,6 +63,44 @@ const propTypes = {
...
+    navElementShape,
+    {
+      children: PropTypes.arrayOf(PropTypes.shape(
+      navElementShape,
+        {
+          children: PropTypes.array,
```
alex-bezek: do we want to limit them to this many children? Could do this recursively
https://stackoverflow.com/questions/32063297/can-a-react-prop-type-be-defined-recursively

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -62,6 +63,44 @@ const propTypes = {
...
+        },
+      )),
+    },
+  )),
+  /**
+   * A unique id set to the help button that will be referred in help menu/popup .
+   */
+  helpId: PropTypes.string.isRequired,
```
alex-bezek: this doesn't seem to be used

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: SO, prop-types here should be array and component will verify the props?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  togglePopup() {
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  render() {
+    const { help, id, ...customProps } = this.props;
+    const helpButton = (
+      <button
```
alex-bezek: should we be using terra buttons?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import Arrange from 'terra-arrange';
+import ResponsiveElement from 'terra-responsive-element';
+import NavHelpContent from './NavHelpContent';
+import IconInfo from '../../icons/IconInfo';
+import Modal from '../modal/Modal';
+import Popup from '../popup/Popup';
```
alex-bezek: This is confusing that we have both 
`import Popup from '../popup/Popup';` in this component, but also 
`import Popup from 'terra-popup';` https://github.com/cerner/terra-consumer/pull/6/files#diff-328c90881807f447a8ea18e6f23ab9f1R7
I would think we would want to distinguish between them. If we have our own custom popup, lets name it as such

Mahmud-Khan: For now replaced Popup with TerraPopup where Popup from terra-core is used. Based on the list of future components terra-consumer will have its own Popup. We can replace TerraPopup with that new one and remove the current custom Popup with a custom header.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Updated with terra button.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,166 @@
...
+        }
+      </Popup>);
+
+    return (
+      <div {...customProps}>
+        <Arrange
+          className={cx('profile')}
+          fitStart={<svg className={cx('profile-icon')}>{profile.avatar}</svg>}
+          fill={<div style={{ paddingLeft: '10px' }}>{profile.name}</div>}
+          fitEnd={<a id="profile-link-icon" onClick={this.openPopup}><svg className={cx('profile-icon')} role="button">{profile.icon}</svg></a>}
```
Mahmud-Khan: Removed the SVG element tag and user terra-icon.

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,93 @@
...
+              <div key={`${element.text}`} className={cx('help-subitem', 'toggler-content-padding')}>{element.text}</div>
+              ))
+            }
+          </Toggler>
+        </div>
+      </button>
+      )
+      :
+      (
+        <button key={`${content.text}`} onClick={() => { window.location = content.uri; }} href="content.uri" className={i > 0 ? cx('help-item', 'help-item-border') : cx('help-item')} >
```
KevinJJackson: use terra-buttons?
you shouldn't need this onclick if the href is set correctly:
`... href={content.uri} ...`

packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx
```diff
@@ -0,0 +1,41 @@
...
+const defaultProps = {
+  linkItems: [],
+};
+
+const ProfileLinks = ({
+  linkItems,
+  ...customProps
+}) => {
+  const linkElements = linkItems.map((linkItem) => {
+    const linkElement = (<button key={`${linkItem.text}`} href={linkItem.uri} onClick={() => { window.location = linkItem.uri; }} className={cx('link', 'nav-item-border')}>
```
KevinJJackson: can just return the element directly.
terra-buttons?
shouldn't need onClick.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import Arrange from 'terra-arrange';
+import ResponsiveElement from 'terra-responsive-element';
+import IconEllipses from 'terra-icon/lib/icon/IconEllipses';
+import Popup from 'terra-popup';
+import Modal from '../modal/Modal';
+import ProfileLinks from './ProfileLinks';
+import navElementShape from '../../NavPropShapes';
```
KevinJJackson: np: keep imports alphabetized within each group (absolute and relative imports)

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+  signoutUrl: PropTypes.string.isRequired,
+  /**
+   * The content of the each profile items.
+   */
+  profileLinks: PropTypes.arrayOf(PropTypes.shape(
+    navElementShape)),
+  /**
+   * An array of items to be displayed as help modal.
+   */
+  help: PropTypes.arrayOf(PropTypes.shape(
```
KevinJJackson: as discussed, help should no longer be a part of the profile popup

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+      <div {...customProps} className={cx('profile')}>
+        <button className={cx('popup-button')}  onClick={() => this.togglePopup()}>
+          <Arrange
+            fitStart={<svg className={cx('icon')}>{avatar}</svg>}
+            fill={<div className={cx('profile-text-padding')}>{name}</div>}
+            fitEnd={<svg className={cx('icon')} id="profile-link-button"><IconEllipses /></svg>}
+            align="center"
+          />
+        </button>
+        <ResponsiveElement responsiveTo="window" defaultElement={defaultElement} medium={large} large={large} />
```
KevinJJackson: just set medium attribute, not large

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -0,0 +1,81 @@
...
+    background: inherit;
+    border: 0;
+    cursor: pointer;
+    font-size: 16px;
+    padding: 20px 25px;
+    text-align: center;
+    width: inherit;
+ 
+    @media screen and (max-width: $terra-medium-breakpoint) {
+      color: var(--terra-consumer--mobile-badge-text-color, #fff);
```
KevinJJackson: should not be reusing a variable here, each element should be themed using a different var.

packages/terra-consumer-nav/tests/jest/Nav.test.jsx
```diff
@@ -16,6 +16,30 @@ const testData = {
     altText: 'test',
     isCard: false,
   },
+  helpItems: [{
+    text: 'Technical Questions',
+    uri: 'http://localhost:8083/',
```
KevinJJackson: should probably keep these consistent with tests and use 8080

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+
+  toggleModal() {
+    this.setState({ isOpen: false, showModal: !this.state.showModal });
+  }
+
+  render() {
+    const { name, avatar, signoutUrl, profileLinks, help, ...customProps } = this.props;
+
+    const signout = (
+      <button
```
KevinJJackson: use terra-button

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+        className={cx('link', 'signout-border')}
+        href={signoutUrl}
+        onClick={() => { window.location = signoutUrl; }}
```
KevinJJackson: shouldnt need this onClick

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+      >
+        <div className={cx('link-text')}>
```
KevinJJackson: set this to the text attribute of the terra-buttons and apply the colors/properties to the button class in your scss.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,146 @@
...
+        content={<div><ProfileLinks linkItems={profileLinks} />
+          {signout}</div>}
+        closeModal={this.togglePopup}
+      />);
+
+    const large = (
+      <Popup
+        isOpen={this.state.isOpen}
+        onRequestClose={this.togglePopup}
+        targetRef={() => document.getElementById('profile-link-button')}
```
KevinJJackson: should we be hard-coding this id, or pass it as a prop, similarly to the help button

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: The children of children are ignored based on initial design discussion with Mike. We can discuss this issue of how many hierarchical levels we want to go. The help modal content list is now designed based on assumption that there will not be children of children and if that assumption is changed need to change that component also. For now keeping it as current form. If we decide not limiting children will refactor the code.

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: Based on a previous code review comment added the prop shapes both on component and sub components.

packages/terra-consumer-nav/package.json
Mahmud-Khan: Removed duplicate.

packages/terra-consumer-nav/package.json
Mahmud-Khan: Modified to keep alphabetic order.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Fixed this issue. This is the current behavior in chrome
screen shot 2017-08-15 at 12 39 03 pm

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Set medium break point only.

packages/terra-consumer-nav/src/Nav.scss
Mahmud-Khan: Reverted the Nav UI changes.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
Mahmud-Khan: Added terra-i18n for internationalization in Layout component.

packages/terra-consumer-nav/src/Nav.scss
Mahmud-Khan: As mention above making the profile background inherit from gradient causes the profile background to look different from the Nav background. It works as expected if the background is not gradient. Making the profile background transparent make the profile to have continuous gradient background but it causes an issue for scrolling. For now made profile background transparent for desktop and inherit from nav for mobile. If anyone has suggestion I appreciate.

packages/terra-consumer-nav/tests/jest/Nav.test.jsx
Mahmud-Khan: Updated.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
Mahmud-Khan: Updated

packages/terra-consumer-nav/src/Nav.jsx
Mahmud-Khan: removed unused prop.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Updated.

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
Mahmud-Khan: Updated.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: removed help for the profile component.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Removed onClick.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Updated with terra-button.

packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx
Mahmud-Khan: Updated

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Updated.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Added prop in nav to send profile pop button id.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
KevinJJackson: should the entire box have a tab-index to begin with? seems like strange behavior and should instead jump straight to the close button, not the popup itself.

Mahmud-Khan: Checked terra-core doc site the entire box has a tab index and after the popup is open the first pressing of tab button focus the popup window.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+  render() {
+    const { name, avatar, id, signoutUrl, profileLinks, intl, ...customProps } = this.props;
+
+    const signout = (
+      <Button
+        className={cx('link', 'signout-border')}
+        href={signoutUrl}
+      >
+        <div>
```
KevinJJackson: np: unneeded div

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
KevinJJackson: np: extra white space in classes

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -3,10 +3,12 @@
...
     z-index: 1000;
```
KevinJJackson: this was a typo on my end and has a weird interaction with your help modal on mobile view, change this to 100 so it's behind your modal/overlay.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
Mahmud-Khan: Updated

packages/terra-consumer-nav/src/Nav.scss
Mahmud-Khan: Updated

packages/terra-consumer-nav/src/components/nav-toggler/NavToggler.scss
```diff
@@ -18,4 +18,9 @@
   .toggler {
     padding: 0;
   }
+
+  .icon {
```
Mahmud-Khan: Added when I was using your previous toggler. Removed the class.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  .link-text-style {
+    text-decoration: none;
+  }
+
+  .popup-content {
+    border-radius: 5px;
+    bottom: 0;
+    box-shadow: 0 1px 5px #666;
```
Mahmud-Khan: Made them consistent.

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,113 @@
...
+      classNameContent={cx('popup-content')}
+      contentAttachment="middle center"
+      popupContent={popupContent}
+    />);
+
+    return (
+
+      <div {...customProps}>
+        <ResponsiveElement responsiveTo="window" defaultElement={defaultElement} medium={popup} />
+        { (this.state.isOpen === false) && helpButton}
```
Mahmud-Khan: Updated.

mhemesath:
@Mahmud-Khan  you need to resolve these conflicts before we can review your branch

packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ChangeLog
 =========
 
-# 0.1.0-BETA.1 - (August 14, 2017)
```
KevinJJackson: don't rewrite the changelog, just add a new entry

Mahmud-Khan: Fixed it.

mhemesath:
A few questions

* When I click help, the text turns white. It shouldn't do that
* Why is there extra empty space in the popup when profile settings is clicked?

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -5,10 +5,11 @@ import Button from 'terra-button';
 import IconClose from 'terra-icon/lib/icon/IconClose';
 import NavItems from './components/nav-items/NavItems';
 import NavLogo from './components/nav-logo/NavLogo';
-import NavProfile from './components/nav-profile/NavProfile';
+import navElementShape from './NavPropShapes';
```
alex-bezek: Whats the difference between navElementShape and navItemShape? The naming is pretty hard to tell any significant difference. It appears navItems can have badges and active status?

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -62,6 +63,30 @@ const propTypes = {
...
+   * Avatar to be displayed in user profile in navigation.
+   */
+  avatar: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  /**
+   * A unique id set to the profile popup button that will be referred in profile popup.
+   */
+  profileId: PropTypes.string.isRequired,
```
alex-bezek: Is there a reason this id needs to be exposed to the consumer? Could it just be something namespaced and unique inside this component?

packages/terra-consumer-nav/src/components/modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Required callback function for use by parent component to update state.
+   */
+
+  closeModal: PropTypes.func.isRequired,
+};
+
+const defaultProps = {
+  title: null,
+  content: null,
```
alex-bezek: Just curious if this being defaulted to null has some significance or if they could be left as undefined?

packages/terra-consumer-nav/src/components/modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+ content,
+ isModalOpen,
+ closeModal,
+ ...customProps
+}) => {
+  const modalHeader = (
+    <div className={cx('modal-header')}>
+      <p className={cx('modal-title')}>
+        {title}
+        <button className={cx('close-button')} onClick={() => closeModal()}><IconClose className={cx('close-icon')} /></button>
```
alex-bezek: any reason we aren't using the terra button?

alex-bezek: Since you are just taking the prop `closeModal` function and passing it to the button component, do you actually need to declare a new anonymous function to pass down? Does it handle the binding or something?

packages/terra-consumer-nav/src/components/modal/Modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const modal = (
+    <Overlay isOpen={isModalOpen} backgroundStyle="dark" className={cx('overlay')}>
+      {modalHeader}
+      <div className={cx('modal')}>
+        {content}
+      </div>
+    </Overlay>);
+
+  return (
+    <div {...customProps}>
```
alex-bezek: do we need an extra div here, or could you just return the overlay as the root element and forgoe the modal const? I'm assuming then you would just pass the custom props to the overlay itself

packages/terra-consumer-nav/src/components/modal/Modal.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  .modal-title {
+    color: var(--terra-consumer--modal-title-color, #fff);
+    font-size: 20px;
+    line-height: 30px;
+    text-align: center;
+  }
+  
+  .overlay {
+    background-color: var(--terra-consumer--overlay-background, #64696c);
```
alex-bezek: any reason we don't just make the terra overlay allow this to be themed? If just for speed and time, can we make an issue and leave a comment in this code to that issue?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,113 @@
...
+import Popup from '../popup/Popup';
+import styles from './NavHelp.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * An array of items to be displayed as help menu/popup.
+   */
+  help: PropTypes.arrayOf(PropTypes.shape(
```
alex-bezek: nit: helpLinks? helpNavs?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,113 @@
...
+        className={cx('nav-help')}
+      >
+        <Arrange
+          fill={<div className={cx('icon')}><IconInfo /></div>}
+          fitEnd={<div className={cx('button-text-padding')}><FormattedMessage id="nav_help" /></div>}
+          align="stretch"
+        />
+      </Button>);
+
+    const popupContent = <NavHelpContent helpContent={help} />;
```
alex-bezek: if you are just passing `help` directly to this component, why are we checking propTypes here and there?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,113 @@
...
+    />);
+
+    const popup = (<Popup
+      title={intl.formatMessage({ id: 'nav_help' })}
+      hasHeader
+      isOpen={this.state.isOpen}
+      targetRef={() => document.getElementById(id)}
+      closePopup={this.togglePopup}
+      contentWidth="320"
+      contentHeight="240"
```
alex-bezek: what happens if we have a bunch of links in the help popup?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -0,0 +1,105 @@
...
+:local {
+  .nav-help {
+    background-color: var(--terra-consumer--help-button-background, #fff);
+    border: 0;
+    border-radius: 20px;
+    box-shadow: 0 1px 1px #666;
+    color: var(--terra-consumer--help-button-color, #3d3d3d);
+    display: inline-block;
+    font-family: inherit;
+    font-size: 16px;
```
alex-bezek: this may be an enhancement for later, but what do you think about making all these font-size's and line-heights either a single variable, or multiple variables?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,94 @@
...
+  handleToggle() {
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  render() {
+    const { helpContent, ...customProps } = this.props;
+    const toggleIcon = this.state.isOpen ? <IconChevronUp className={cx('icon')} /> : <IconChevronDown className={cx('icon')} />;
+    const contentList = helpContent.map((content, i) => {
+      const contentElement = content.children.length > 0 ?
+      (<button key={`${content.text}`} onClick={this.handleToggle} className={i > 0 ? cx('help-item', 'help-item-border') : cx('help-item')}>
```
alex-bezek: any reason to not use the terra button?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,94 @@
...
+        <div className={cx('help-item-text')}>
+          <Arrange
+            align="stretch"
+            fitStart={<div className={cx('icon')}>{content.icon}</div>}
```
alex-bezek: according to your prop type shape, icons can be strings right? Is it expecting a url to an aws icon, or like an inline svg content?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,94 @@
...
+            fill={<div className={cx('item-text-padding')}>{content.text}</div>}
+            fitEnd={<div className={cx('icon')}>{toggleIcon}</div>}
+          />
+          <Toggler isOpen={this.state.isOpen} isAnimated className={cx('toggler-padding')}>
+            { content.children.map(element => (
+              <div key={`${element.text}`} className={cx('help-subitem', 'toggler-content-padding')}>{element.text}</div>
```
alex-bezek: if you are just expecting text, i would think spans of paragraphs would be more semantic

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,94 @@
...
+              ))
+            }
+          </Toggler>
+        </div>
+      </button>
+      )
+      :
```
alex-bezek: IMO this is getting a bit big for a ternary operator

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -0,0 +1,94 @@
...
+      (
+        <Button key={`${content.text}`} href={content.uri} className={i > 0 ? cx('help-item', 'help-item-border') : cx('help-item')} >
+          <div className={cx('help-item-text')}>
+            <Arrange
+              align="center"
+              fitStart={<svg className={cx('icon')}>{content.icon}</svg>}
```
alex-bezek: are we expecting content.icon to be a path object in this scenario?

packages/terra-consumer-nav/src/components/popup/Popup.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import IconClose from 'terra-icon/lib/icon/IconClose';
+import TerraPopup from 'terra-popup';
+import styles from './Popup.scss';
+
+const cx = classNames.bind(styles);
+
+
```
alex-bezek: nit: extra line

packages/terra-consumer-nav/src/components/popup/Popup.jsx
```diff
@@ -0,0 +1,80 @@
...
+const propTypes = {
+  /**
+   * Title to be rendered on top of the popup content
+   */
+  title: PropTypes.oneOfType([
```
alex-bezek: In general, the terra team tries to stray away from prop types that can be multiple very different types

packages/terra-consumer-nav/src/components/popup/Popup.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
alex-bezek: Popup seems very generic

packages/terra-consumer-nav/src/components/popup/Popup.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+const Popup = ({
+ title,
+ hasHeader,
+ popupContent,
+ isOpen,
+ closePopup,
+ ...customProps
+}) => {
+  const popupHeader = (
```
alex-bezek: since you have a prop for conditionally displaying the header, you can make this a lazy evaluated const to not run all this when the condition is false

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,133 @@
...
+      >
+        <FormattedMessage id="nav_signout" />
+      </Button>);
+
+    const defaultElement = (
+      <Modal
+        isModalOpen={this.state.isOpen}
+        title={intl.formatMessage({ id: 'nav_profile_title' })}
+        content={<div><ProfileLinks linkItems={profileLinks} />
+          {signout}</div>}
```
alex-bezek: nit: align your brackets and space this out

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -0,0 +1,133 @@
...
+      );
+
+    return (
+
+      <div {...customProps} className={cx('profile')}>
+        <Button className={cx('popup-button')} onClick={() => this.togglePopup()}>
+          <Arrange
+            fitStart={<svg className={cx('icon')}>{avatar}</svg>}
+            fill={<div className={cx('profile-text-padding')}>{name}</div>}
+            fitEnd={<IconEllipses className={cx('icon')} id={id} />}
```
alex-bezek: this is interesting. can you explain why we need to pass the id to the icon?

alex-bezek:
for the css variable names, I would think we want to scope them to this particular element. With some of the current names, we can never have anything profile or help related in other components now

packages/terra-consumer-nav/src/icons/IconInfo.jsx
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable */
+import React from 'react';
+import IconBase from 'terra-icon/lib/IconBase';
+
+const SvgIcon = (customProps) => {
+  const attributes = Object.assign({}, customProps);
```
alex-bezek: specific reason for the object assign as opposed to just deconstructing and spreading customProps like in other elements?

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -62,6 +34,27 @@ const propTypes = {
...
+   */
+  userName: PropTypes.string.isRequired,
+  /**
+   * Avatar to be displayed in user profile in navigation.
+   */
+  avatar: PropTypes.PropTypes.element,
+  /**
+   * A unique id set to the profile popup button that will be referred in profile popup.
+   */
+  profileId: PropTypes.string.isRequired,
```
alex-bezek: we forgot to chat about this in our call, but I had talked with Yao. He just meant profile and help shouldn't declare their own ID's as there is a chance they could be used more than once on a page somehow. but navigatin is a singleton so it would be fine to statically declare the Id's in this root level component rather than the consumer have to pass them down

packages/terra-consumer-nav/src/components/modal/Modal.scss
alex-bezek: Discussed, @Mahmud-Khan will be logging a issue to terra to do this, and leave a comment in the code here pointing to that issue. We will pick that up when we have time

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
alex-bezek: talked IRL: scroll bar. This is fine for now

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
alex-bezek: +1 makes sense

packages/terra-consumer-nav/src/components/modal/Modal.jsx
alex-bezek: @Mahmud-Khan can you address this issue across other components, not just this one (like nav.jsx)

packages/terra-consumer-nav/src/components/modal/Modal.jsx
alex-bezek: https://github.com/cerner/terra-consumer/pull/6/files#diff-e03eae71444e9381aeb7a87cf8959e28R58
still see another non terra-button component

alex-bezek: addressed

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -0,0 +1,113 @@
...
+        className={cx('nav-help')}
+      >
+        <Arrange
+          fill={<div className={cx('icon')} ><IconOutlineQuestionMark /></div>}
+          fitEnd={<div className={cx('button-text-padding')}><FormattedMessage id="nav_help" /></div>}
+          align="stretch"
+        />
+      </Button>);
+
+    const popupContent = <NavHelpContent helpContent={helpNavs} />;
```
alex-bezek: https://github.com/cerner/terra-consumer/pull/6#discussion_r134125475
this question still isn't addressed

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
alex-bezek: discussed in person. Using just element and the assumption will be that it will always be an svg react component. if url's are provided, a presenter layer can create react components that bootstrap themselves with the url

KevinJJackson:
+1

PR #9, opened 2017-08-11
Updating night watch to look at the node installed version of chrome in order to fix issue #8

### LINKED ISSUE #8: Nightwatch tests have slowed in Travis (and potentially locally)
# Issue Description
Looking at the build here: https://travis-ci.org/cerner/terra-consumer/builds/262669204, the night watch tests have started taking 10 seconds each.  I noticed that happening while running locally as well.  Looking at the logs, this time that is being added happens because it is taking 10 seconds to try and resize the browser and fails.  

It appears there is a newer version of chrome that fixes this issue, but the way we load the chrome driver, we are dependent upon the version that is loaded on the travis boxes.  I have a potential fix that I will include in a PR so that we are referencing the chromedriver npm module, which seems like it has been kept up to date more frequently.  Also, the benefit of that is consistency of runs across machines.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
Builds don't hang when trying to resize the browser and perform reasonably

## Current Behavior
Builds are hanging while trying to resize the browser and are taking 10 seconds per test to run.

ryanthemanuel:
### Summary
In order to fix #8, this PR will update night watch to look at the node installed version of chrome rather than the version of chrome that happens to be installed on the box where the tests are running


ryanthemanuel:
Looks like travis is still taking 10s a test.  I'll do some additional digging.

ryanthemanuel:
Added some new travis config.  This should be good to go now.

mhemesath:
thanks everyone

PR #13, opened 2017-08-18
Navigation ui and docs

KevinJJackson:
### Summary
Updating Navigation Logo to no longer have a width constraint on mobile/tablet.
Added new feature on nav-toggler to open the first toggler found with an 'active' subItem or close all if none is found.
Added doc site for nav.

@cerner/terra


alex-bezek:
Mike found a minor issue you might be able to fix here.
It seems when you turn card on, you don't get the box shadow.
screen shot 2017-08-22 at 8 36 15 am
But if you turn overflow hidden off the parent container, you do get it
screen shot 2017-08-22 at 8 36 23 am

KevinJJackson:
@alex-bezek @mhemesath I can do that here. I also added 10px between the card and the logo so the max logo size while in card (to look nice) is 250x150. If that's an issue, I can remove that padding for now as well.

image

PR #17, opened 2017-08-24
Nav cleanup

KevinJJackson:
### Summary
@mhemesath @alex-bezek @Mahmud-Khan 

Added
- Icon support for navigation items.
- target attribute support for navigation items.

Changed
- Profile settings modal now opens as expected in safari tablet mode.
- Minor UI fixes.
- PropType consistency for links/paths.

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -4,10 +4,13 @@ ChangeLog
 # 0.1.0-BETA.3 - (--)
 
 ### Added
+- Icon support for navigation items.
+- target attribute support for navigation items.
 
 ### Changed
-
-### Removed
+- Profile settings modal now opens as expected in safari tablet mode.
+- Minor UI fixes.
+- PropType consistency for links/paths.
 
 -----------------
 
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,7 +1,7 @@
 ChangeLog
 =========
 
-# 0.1.0-BETA.3 - (--)
+# 0.1.0-BETA.3 - (August 25, 2017)
 
 ### Added
 - Icon support for navigation items.
@@ -14,7 +14,7 @@ ChangeLog
 
 -----------------
 
-# 0.1.0-BETA.2 - (August 24, 2017)
+# 0.1.0-BETA.2 - (August 22, 2017)
 
 ### Added
 - nav-help: Component that renders a help button in bottom right corner of the screen.
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,6 +1,16 @@
 ChangeLog
 =========
 
+# 0.1.0-BETA.3 - (--)
+
+### Added
+
+### Changed
+
+### Removed
+
+-----------------
+
 # 0.1.0-BETA.2 - (August 24, 2017)
 
 ### Added
--- a/packages/terra-consumer-nav/src/Nav.jsx
+++ b/packages/terra-consumer-nav/src/Nav.jsx
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Button from 'terra-button';
 import IconClose from 'terra-icon/lib/icon/IconClose';
+import TerraPopup from 'terra-popup';
+import ResponsiveElement from 'terra-responsive-element';
 import NavItems from './components/nav-items/NavItems';
+import Modal from './components/modal/Modal';
 import NavLogo from './components/nav-logo/NavLogo';
 import QuickLink from './components/quick-links/QuickLink';
 import QuickLinks from './components/quick-links/QuickLinks';
@@ -16,9 +19,7 @@ const propTypes = {
   /**
    * An array of objects to be displayed as quick link options.
    */
-  quickLinks: PropTypes.arrayOf(PropTypes.shape({
-    text: PropTypes.string.isRequired,
-  })),
+  quickLinks: PropTypes.array,
   /**
    * An array of objects to be displayed as nav link options.
    */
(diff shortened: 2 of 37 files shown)


packages/terra-consumer-nav/src/components/modal/Modal.jsx
```diff
@@ -1,3 +1,4 @@
+
```
mhemesath: Why do we have our own modal nested in consumer-nav?

packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
```diff
@@ -21,6 +21,15 @@
     }
   }
 
+  .nav-item-icon svg {
```
mhemesath: this wont' suffice for font icons... will need to set a font-size to control those

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -27,12 +27,12 @@ const defaultProps = {
...
 
 const NavLogo = ({
-  path,
+  url,
   altText,
   isCard,
   ...customProps
 }) => {
-  const image = <img className={cx('img')} src={path} alt={altText} />;
+  const image = <img className={cx('img')} src={url} alt={altText} />;
```
mhemesath: I'm actually wondering if we shuld just have the navLogo be a react component.... Then all the link, card, and other logic can be handled outside of the nav framework.

packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,6 +1,19 @@
 ChangeLog
 =========
 
+# 0.1.0-BETA.3 - (--)
```
alex-bezek: can we make sure to add this date of whatever day you commit? Otherwise I have to keep updating it every time i release

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -16,9 +19,7 @@ const propTypes = {
   /**
    * An array of objects to be displayed as quick link options.
    */
-  quickLinks: PropTypes.arrayOf(PropTypes.shape({
-    text: PropTypes.string.isRequired,
-  })),
```
alex-bezek: why was this removed?

packages/terra-consumer-nav/src/components/modal/Modal.jsx
KevinJJackson: this is definitely feasible and makes sense to me. I can do that separately from this PR if that's better.

packages/terra-consumer-nav/CHANGELOG.md
KevinJJackson: yeah that's my bad, i thought the release date was supposed to go here since releases may not always be the same day as the commit. I'll update this

packages/terra-consumer-nav/src/Nav.jsx
KevinJJackson: it wasn't removed, just changed. I just removed the validation from the nav to leave the more specific validation to the component.

packages/terra-consumer-nav/src/components/modal/Modal.jsx
alex-bezek: I think this is the master plan in the end (same with card and popup). This is here for now for speed

packages/terra-consumer-nav/src/NavPropShapes.js
```diff
@@ -6,14 +6,18 @@ const navItemShape = {
    */
   text: PropTypes.string.isRequired,
   /**
-   * The path the profile link would redirect to.
+   * The path the link would redirect to.
    */
-  uri: PropTypes.string.isRequired,
+  url: PropTypes.string,
```
KevinJJackson: oops, removed this when i was debugging a weird issue, will put it back, thanks for pointing it out!

packages/terra-consumer-nav/src/components/modal/Modal.jsx
```diff
@@ -14,50 +15,44 @@ const propTypes = {
    */
   title: PropTypes.string,
   /**
-   * Content to be rendered in the body of the dialog box.
-   */
-  content: PropTypes.element,
```
alex-bezek: is most of this just an attempt to align us with the terra modal more closely?
https://github.com/cerner/terra-core/blob/master/packages/terra-modal/src/Modal.jsx

KevinJJackson: for the most part, figured some consistency would be nice. also removed some unneeded nested classes from the title.

packages/terra-consumer-nav/src/NavPropShapes.js
Mahmud-Khan: There may be a case where the url is not needed. Like clicking a navigation item expand a sub item like the following.
screen shot 2017-08-24 at 3 20 59 pm

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
alex-bezek: Are you saying remove this completely and let the consumer provide this as a prop to `Nav` or to just export this module for use by the consumer?

packages/terra-consumer-nav/src/NavPropShapes.js
KevinJJackson: ah you are correct here, we were previously bypassing the required field by simply passing `url: '',` into the component. 
example: https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/tests/nightwatch/DefaultNav.jsx#L37

packages/terra-consumer-nav/CHANGELOG.md
alex-bezek: I think you are technically correct. But i'm not sure the date matters that much. I'll work with @bjankord to see what his process is to make things easiest

packages/terra-consumer-nav/src/Nav.jsx
alex-bezek: i added just the text validation here because text is used below for the key. The idea is if a component takes props and uses some value off them, it should declare the type for that prop's value.

packages/terra-consumer-nav/src/NavPropShapes.js
alex-bezek: thats a good point @Mahmud-Khan . Not required makes sense.

packages/terra-consumer-nav/src/Nav.jsx
KevinJJackson: right, makes sense, will revert this change

PR #21, opened 2017-08-29
Nav ui uplift

KevinJJackson:
### Summary
Changes made to profile link to allow for signin as well as ui fixes across the navigation component.

### Added
- Profile now has option to display sign-in link.
- Profile now overlays items on long navigation lists.

### Changed
- Minor UI fixes.
- Font antialiased fixes for chrome/firefox.
- Hover effects for nav-items (mobile/desktop)
- Scrolling content on help popup no longer scrolls title.
- Removed id as prop requirement for help/profile.
- Multiple help togglers do not open simultaneously, although there can be more than one open at a time.
@cerner/terra

https://terra-consumer-pr-21.herokuapp.com/static/#/tests/nav-tests/default

### DIFF
--- a/packages/terra-consumer-nav/src/Nav.jsx
+++ b/packages/terra-consumer-nav/src/Nav.jsx
@@ -19,9 +19,11 @@ const propTypes = {
   /**
    * An array of objects to be displayed as quick link options.
    */
-  quickLinks: PropTypes.arrayOf(PropTypes.shape({
-    text: PropTypes.string.isRequired,
-  })),
+  quickLinks: PropTypes.arrayOf(
+    PropTypes.shape({
+      text: PropTypes.string.isRequired,
+    }),
+  ),
   /**
    * An array of objects to be displayed as nav link options.
    */
--- a/packages/terra-consumer-nav/src/Nav.jsx
+++ b/packages/terra-consumer-nav/src/Nav.jsx
@@ -28,22 +28,11 @@ const propTypes = {
    * An array of objects to be displayed as nav link options.
    */
   navItems: PropTypes.array.isRequired,
-  /**
-   * An array of nav items to be displayed on the user profile/ settings menu/popup.
-   */
-  profileLinks: PropTypes.array.isRequired,
-  /**
-   * User name to be displayed in the profile in navigation.
-   */
-  userName: PropTypes.string.isRequired,
-  /**
-   * Avatar to be displayed in user profile in navigation.
-   */
-  avatar: PropTypes.PropTypes.element,
-  /**
-   * The path signout button would redirect to.
-   */
-  signoutUrl: PropTypes.string.isRequired,
+
+  profile: PropTypes.shape({
+    signinUrl: PropTypes.string,
+    avatar: PropTypes.element,
+  }),
   /**
    * An object defining the logo to be displayed
    */
@@ -74,8 +63,7 @@ const propTypes = {
 const defaultProps = {
   quickLinks: [],
   navItems: [],
-  profileLinks: [],
-  avatar: null,
+  profile: {},
   logo: {},
 };
 
@@ -110,8 +98,10 @@ class Nav extends React.Component {
   }
 
   render() {
-    const { quickLinks, navItems, profileLinks, userName, avatar, signoutUrl, logo, isMobileNavOpen, onRequestClose, ...customProps } = this.props;
+    const { quickLinks, navItems, profile, logo, isMobileNavOpen, onRequestClose, ...customProps } = this.props;
     const profileId = 'profile-popup-button';
+
+    const willRenderProfile = profile.userName || profile.avatar || profile.signinUrl;
 
     const defaultElement = (
       <Modal
@@ -138,7 +128,7 @@ class Nav extends React.Component {
     );
 
     return (
-      <div className={cx('nav')}>
+      <div id="terra-consumer-nav">
         <div {...customProps} className={cx('nav', customProps.className)} aria-hidden={!isMobileNavOpen}>
           <Button icon={<IconClose />} className={cx('close-button')} onClick={() => { onRequestClose(); }} />
           <NavLogo {...logo} />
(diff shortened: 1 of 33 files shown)


packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -108,9 +98,11 @@ class Nav extends React.Component {
   }
 
   render() {
-    const { quickLinks, navItems, profileLinks, userName, avatar, signoutUrl, logo, isMobileNavOpen, onRequestClose, ...customProps } = this.props;
+    const { quickLinks, navItems, profile, logo, isMobileNavOpen, onRequestClose, ...customProps } = this.props;
     const profileId = 'profile-popup-button';
 
+    const willRenderProfile = profile.userName || profile.avatar || profile.signinUrl;
```
alex-bezek: do we have some use case where we aren't rendering the profile?

KevinJJackson: yes we had this as a new requirement:
`✓ - if sign-in url provided and avatar/user NOT provided, Profile link should just be a sign-in link. 
✓ -- If no sign-in url then do not render component at all `

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -12,35 +12,46 @@ const cx = classNames.bind(styles);
...
     super();
-    this.state = ({ isOpen: false });
+    this.state = ({ openTogglers: [] });
 
     this.handleToggle = this.handleToggle.bind(this);
   }
 
-  handleToggle() {
-    this.setState({ isOpen: !this.state.isOpen });
+  handleToggle(toggleIndex) {
```
alex-bezek: seems like this logic would all be alot simpler if you just had a hash, where the key was the toggleIndex, and the value was a boolean for isOpen. then handleToggle would be
```
state = {
  togglers: {
     1: true,
     2: false,
     3: false,
  }
}
handleToggle(toggleIndex){
   this.setState({
      togglers: {...this.state.togglers, toggleIndex: !this.state.togglers[toggleIndex]}
    })
}
```

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -13,31 +14,40 @@ const cx = classNames.bind(styles);
...
+  /**
    * The content of the each profile items.
    */
-  profileLinks: PropTypes.arrayOf(PropTypes.shape(
-    navElementShape)),
-    /**
-    */
+  profileLinks: PropTypes.arrayOf(
+    PropTypes.shape(
+      navElementShape,
```
alex-bezek: do we even use any of the navElementShape props here? ideally we only check the prop types in the component we are consuming them in. 
In my branch, i'm seeing if i have a single component that handles rendering the link (to handle all the target, type, react router, custom props, etc) if its the only component that needs all these props, then I will remove navElementShape

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -46,37 +56,53 @@ const propTypes = {
...
 const UserProfile = ({
-  name, avatar, id, signoutUrl, profileLinks, handleClick, intl, ...customProps
+  userName, avatar, id, signoutUrl, signinUrl, isSignIn, profileLinks, handleClick, intl, ...customProps
 }) => {
-  const content = (
-    <div>
-      <ProfileLinks linkItems={profileLinks} />
-      <Button className={cx('link', 'signout-border')} href={signoutUrl}>
-        <FormattedMessage id="nav_signout" />
+  let ret;
```
alex-bezek: whats a ret?

packages/terra-consumer-nav/tests/jest/components/UserProfile.test.jsx
```diff
@@ -21,10 +21,20 @@ describe('UserProfile', () => {
   it('should render a user profile with avatar,name and popup button', () => {
     const wrapper = shallow(<UserProfile
       profileLinks={profileLinks}
-      name="Frank Lampard"
+      name="Anthony Martial"
```
alex-bezek: +1

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: yeah these seem to be passed straight through to the child, will remove validation from this.

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
KevinJJackson: you're right, hash is easier here, will update this

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: a shortened form of return when i'm typing fast and not thinking of good var names haha

packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx
```diff
@@ -32,14 +32,15 @@ const propTypes = {
...
-  id: PropTypes.string.isRequired,
+  id: PropTypes.string,
   /**
    * Injected react-intl formatting api
    */
   intl: intlShape.isRequired,
 };
 
 const defaultProps = {
+  id: '--terra-consumer--nav-help-button',
```
zhongyn: any reason using css custom property format here?

KevinJJackson: figured it'd be unique enough to not cause conflicts anywhere. Can change if there is a standard to follow here.

zhongyn: that format is reserved for css custom variables, so i would remove the double dash

zhongyn:
+1

PR #23, opened 2017-08-31
integrate react router and add dex example

zhongyn:
### Summary
1. Remove quick links
2. Add smart link handling external link and react router link
3. Add DEX integration example

There is an issue with test routes and we can see examples via:
Portal: https://terra-consumer-pr-23.herokuapp.com/static/#/tests/nav-tests/portal
Dex: https://terra-consumer-pr-23.herokuapp.com/static/#/tests/nav-tests/dex


packages/terra-consumer-nav/src/NavPropShapes.js
```diff
@@ -22,6 +22,10 @@ export const navItemShape = {
    */
   isActive: PropTypes.bool,
   /**
+   * Whether or not the link is external.
+   */
+  isExternal: PropTypes.bool,
+  /**
    * An optional badge. When supplied, displays the value inline, styled alongside the text.
    */
   badgeValue: PropTypes.number,
```
alex-bezek: @KevinJJackson added a default props section. Should isExternal be added there?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -55,16 +56,22 @@ class NavHelpContent extends React.Component {
...
-            className={cx('help-item-text')}
-            align="center"
-            fitStart={<div className={cx('icon')}>{content.icon}</div>}
-            fill={<div className={cx('item-text-padding')}>{content.text}</div>}
-          />
-        </Button>);
+        contentElement = (
+          <SmartLink
+            className={cx('help-item')}
+            key={`${content.text}`}
```
amf9t2: can we do {content.text}  ?

packages/terra-consumer-nav/src/NavPropShapes.js
zhongyn: it seems the default props section defined in this file is never used. Since isExternal is only used in smartlink, i prefer setting the default there

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
zhongyn: good catch

packages/terra-consumer-nav/tests/nightwatch/nav-spec.js
```diff
@@ -7,7 +7,7 @@ module.exports = resizeTo(['small', 'tiny', 'medium', 'large', 'huge', 'enormous
   '@tags': ['nav'],
   'Displays a default nav': (browser) => {
```
amf9t2: np: 'Display a portal nav'

packages/terra-consumer-nav/src/NavPropShapes.js
KevinJJackson: it is currently not used and was created in advance of the targets. If it's determined they aren't needed and each component should default their own then feel free to remove it, otherwise they should be applied to navItems, profileLinks and navHelp.

I agree that anything specific to a component like isExternal or badgeValue for nav-items should not be in the default structure and should be left to the component.

packages/terra-consumer-nav/tests/nightwatch/NavTests.jsx
```diff
@@ -1,12 +1,13 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
 import React from 'react';
-import { Link } from 'react-router';
 
 const NavTests = () => (
   <div>
     <ul>
-      <li><Link to="/tests/nav-tests/default">Nav - Default</Link></li>
+      test
```
KevinJJackson: np: remove

PR #31, opened 2017-09-05
Update base dependencies

bjankord:
### Summary
Removes base dependency from components which are not using it. Bumps base in the site package to the latest major release.

@cerner/terra


alex-bezek:
@mhemesath does this fit your needs to get nav working with the right base.css file for portal rather than upgrading all our terra dependencies?
https://jira2.cerner.com/browse/PORTALDEV-36251

packages/terra-consumer-icon/package.json
```diff
@@ -33,12 +33,10 @@
   "peerDependencies": {
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
```
alex-bezek: the comma on this line needs to be removed as well @bjankord . Thanks for making this PR to our repo!

mhemesath:
I think we need to update  every terra dependency because they are still referencing the old version of terra base

alex-bezek:
Looks like @amf9t2 is fixing that issue here https://github.com/cerner/terra-consumer/pull/32

bjankord:
We can close this in favor of using @amf9t2's PR #32

PR #32, opened 2017-09-06
upgrading terra-core packages

amf9t2:


packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 ChangeLog
 =========
 
+# 0.1.0-BETA.6 - (September 05, 2017)
```
KevinJJackson: beta.5 hasn't been released yet (https://www.npmjs.com/package/terra-consumer-nav). I believe we should just append these changes instead of listing a new release tag.

amf9t2: [updated](https://github.com/cerner/terra-consumer/pull/32/commits/b5f983ca08d64fdd3c1495887696edff749a47f1)
Thanks.

PR #33, opened 2017-09-06
Punch through update

KevinJJackson:
### Summary
Updated mobile navigation to present the punch through effect as desktop navigation does.
Updated profile bg to match the rest of the nav.

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
@@ -20,7 +20,7 @@
     text-decoration: none;
 
     &:hover {
-      background: rgba(255, 255, 255, 0.4);
+      background: var(--terra-consumer-link-hover-background, rgba(255, 255, 255, 0.4));
       color: var(--terra-consumer-nav-item-color);
       text-decoration: underline;
     }
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
@@ -19,8 +19,7 @@
     display: block;
     text-decoration: none;
 
-    &:hover,
-    &:focus {
+    &:hover {
       background: rgba(255, 255, 255, 0.4);
       color: var(--terra-consumer-nav-item-color);
       text-decoration: underline;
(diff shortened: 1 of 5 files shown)


packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -58,10 +58,6 @@
     text-align: left;
     width: inherit;
 
-    @media screen and (max-width: $terra-medium-breakpoint) {
-      margin-top: 0;
-    }
-
```
alex-bezek: why was this removed?

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -47,7 +47,7 @@
   }
 
   .popup-button {
-    background: var(--terra-consumer-nav-background, rgba(255, 255, 255, 0.5));
+    background: var(--terra-consumer-nav-background, rgba(255, 255, 255, 0.25));
```
alex-bezek: I'm assuming this was correlated with some UX decision?

KevinJJackson: yup, .25 is what the rest of the nav is using as the opacity. The profile link with .5 was drawing too much attention to it.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
KevinJJackson: UX decision to keep the separation between nav items and profile link via the 3px margin-top punch through

packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
```diff
@@ -19,8 +19,7 @@
     display: block;
     text-decoration: none;
 
-    &:hover,
-    &:focus {
+    &:hover {
       background: rgba(255, 255, 255, 0.4);
```
mhemesath: Do these need to be theme-able?

KevinJJackson: probably not a bad idea to be, i will update the hover effects to be theme-able

PR #34, opened 2017-09-06
added nav burger component

KevinJJackson:
### Summary
Extracted Nav 'burger' menu icon into a button component.
Updated layout to use Nav.Burger


packages/terra-consumer-layout/src/Layout.jsx
```diff
@@ -56,7 +55,7 @@ class Layout extends React.Component {
             onRequestClose={this.toggleNav}
           />
           <main id="main-container" className={cx('main-container', this.state.isMobileNavOpen && 'nav-open')}>
-            <button className={cx('nav-burger')} onClick={this.toggleNav}><IconMenu /></button>
+            <Nav.Burger className={cx('nav-burger')} handleClick={this.toggleNav} />
```
mhemesath: haha can we call this toggler instead

mhemesath: You can leave it if we can call something fries?

packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.scss
```diff
@@ -0,0 +1,24 @@
...
+:local {
+  .burger {
+    background: var(--terra-consumer-burger-background, none);
+    border: var(--terra-consumer-burger-border, 0);
+    box-shadow: 0;
+    color: var(--terra-consumer-burger-color, #fff);
+    display: block;
+    font-size: 20px;
+    padding: 0;
+    position: absolute;
```
mhemesath: These shouldn't be part of the toggler but rather the component that wraps it (which would be part of layout).

packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.scss
```diff
@@ -0,0 +1,24 @@
...
+    @media screen and (min-width: $terra-medium-breakpoint) {
+      display: none;
```
mhemesath: These shouldn't be part of the toggler but rather the component that wraps it (which would be part of layout).

packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.jsx
```diff
@@ -0,0 +1,29 @@
...
+  handleClick: PropTypes.func.isRequired,
+};
+
+const NavBurger = ({
+  handleClick,
+  ...customProps
+}) => (
+  <Button
+    {...customProps}
+    onClick={handleClick}
```
mhemesath: question? Should we do this or is it fine if they just pass onClick into customProps?

alex-bezek: this would throw a prop warning to the consumer making sure they have a handleClick method. I think the terra team tries to not use prop's that are the same name as actual html props.

packages/terra-consumer-layout/src/Layout.jsx
KevinJJackson: haha i really liked typing out nav burger the whole time, I vote we find something to call fries

zhongyn: +1burger

alex-bezek:
cutting a beta. Please update the chanelog to reflect the next beta

PR #35, opened 2017-09-06
updated terra dependencies except terra-toolkit

WhiteLight1107:
Updated all terra-dependencies except terra-toolkit, this will be done at a later point


WhiteLight1107:
already done here https://github.com/cerner/terra-consumer/pull/32

PR #36, opened 2017-09-06
removed nav-container-width variable

WhiteLight1107:
Removed unneeded terra-consumer--nav-container-width variable


alex-bezek:
cutting a beta. Please update the chanelog to reflect the next beta

PR #37, opened 2017-09-06
fix IE nav 100% height

zhongyn:
### Summary
Jira: https://jira2.cerner.com/browse/PORTALDEV-36253
On IE nav height doesn't stretch to the window height:
screen shot 2017-09-05 at 2 20 06 pm

### DIFF
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -27,6 +27,7 @@
     background: var(--terra-consumer-nav-background, rgba(255, 255, 255, 0.25));
     color: var(--terra-consumer-nav-text-color);
     height: 100%;
+    min-height: 100vh;
     overflow-y: auto;
     padding-bottom: 60px;
     padding-top: var(--terra-consumer-nav-padding-top, 25px);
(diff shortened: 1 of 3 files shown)


packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -27,6 +27,7 @@
     background: var(--terra-consumer-nav-background, rgba(255, 255, 255, 0.25));
     color: var(--terra-consumer-nav-text-color);
     height: 100%;
```
mhemesath: shoudln't height and min-height be the same unit value?

zhongyn: since nav is relative positioned, (height:100vh) gives nav a fixed height equal to the window height. it doesn't work when the body content is longer than window height. see screenshot:
screen shot 2017-09-06 at 3 18 35 pm

mhemesath: ok thanks that makes sense. Maybe leave that as a comment in the code

packages/terra-consumer-site/webpack.config.js
```diff
@@ -15,7 +15,7 @@ const rtl = require('postcss-rtl');
 module.exports = {
   entry: {
     'babel-polyfill': 'babel-polyfill',
-    'terra-core': path.resolve(path.join(__dirname, 'src', 'Index')),
+    'terra-consumer': path.resolve(path.join(__dirname, 'src', 'Index')),
```
alex-bezek: is this just something we missed in our initial setup?

zhongyn: i think so

PR #39, opened 2017-09-06
logo height fix - matching to parent height

amf9t2:


packages/terra-consumer-nav/src/components/nav-logo/NavLogo.scss
```diff
@@ -17,6 +17,7 @@
     color: #3d3d3d;
     display: block;
     margin: auto;
+    max-height: 120px;
```
mhemesath: can we remove max-height from other container?

amf9t2: good point removed  form parent.

I have question w.r.t callouts 
image
with this change we are overwriting the height to 120 px max. The strange part is callout doesn't have specific logo height for mobile so not sure from where 120 px came in. Or the callout may be outdated. For now I am keeping max height to 120 px and removing the height constraint from parent container (media query).

amf9t2:
logo_height

PR #40, opened 2017-09-06
Help button fix

Mahmud-Khan:
### Summary

- Fixed the help-button alignment issue with. The help button would stick to the bottom if the content does not fill the screen. But if the content height is more than viewport height the help-button would be below the content.
help_button_fix_static

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/CHANGELOG.md
+++ b/packages/terra-consumer-layout/CHANGELOG.md
@@ -2,14 +2,15 @@ ChangeLog
 =========
 
 
-.6 Beta
+# 0.1.0-BETA.2 - (September 07, 2017)
 -----------------
 Initial stable release
-9/6/17:
- - Removed uneeded terra-consumer--nav-container-width variable
+### Changed
+ - Fixed the help-button alignment issue.
+ - Removed unused terra-consumer--nav-container-width variable
  - Added themeable padding for main container
+
 
 # 0.1.0-BETA.1 - (September 06, 2017)
 -----------------
 Initial beta release for the layout component
-
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -5,7 +5,7 @@
   --terra-consumer-nav-active-item-background: var(--terra-consumer-body-background);
   --terra-consumer-nav-item-color: #fff;
 
-  --terra-consumer-nav-width: 300px;
+  --terra-consumer-nav-width: 320px;
   --terra-consumer-nav-text-color: var(--terra-consumer-nav-item-color);
 
   --terra-consumer-profile-link-width: var(--terra-consumer-nav-width);
@@ -56,8 +56,12 @@
   }
 
   .help-button {
-    margin: 15px;
+    padding: 20px;
+    position: fixed;
+    position: sticky;
+    right: 0;
     text-align: right;
+    top: 90%;
 
     @media screen and (min-width: $terra-medium-breakpoint) {
       bottom: 15px;
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -36,7 +36,7 @@
 
   .main-content {
     @media screen and (max-width: $terra-medium-breakpoint) {
-      min-height: calc(100vh - 40px);
+      min-height: calc(100vh - 120px);
     }
   }
 
@@ -62,11 +62,8 @@
   }
 
   .help-button {
-    bottom: 0;
-    padding: 20px;
-    position: fixed;
-    position: sticky;
-    right: 0;
+    margin: 15px;
+    position: static;
     text-align: right;
 
     @media screen and (min-width: $terra-medium-breakpoint) {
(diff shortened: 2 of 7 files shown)


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -55,8 +55,12 @@
   }
 
   .help-button {
-    margin: 15px;
+    padding: 20px;
+    position: fixed;
+    position: sticky;
+    right: 0;
     text-align: right;
+    top: 90%;
```
mhemesath: this will be variable depending on the height of the screen.... why not use bottom instead?

Mahmud-Khan: Right now the help-button is part of the main container. Used bottom keeping the button inside the main container. Since the sticky position is relative to the parent component the button moves to the top when there is no content like the following example.
screen shot 2017-09-06 at 4 48 59 pm

Bottom works in we move the help-button outside the main container. But in that case, the help button is not relative to the content of the main container.

mhemesath: help button should stay at the bottom of the page I think

packages/terra-consumer-layout/CHANGELOG.md
```diff
@@ -4,3 +4,6 @@ ChangeLog
 # 0.1.0-BETA.1 - (September 06, 2017)
 -----------------
 Initial beta release for the layout component
+
+### Changed
+- Fixed the help-button alignment issue.
```
alex-bezek: I released a new beta today, so i think this needs to be part of a new beta tag in the changelog

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -55,8 +55,12 @@
   }
 
   .help-button {
-    margin: 15px;
+    padding: 20px;
+    position: fixed;
+    position: sticky;
```
alex-bezek: how are we defining postion twice? Sorry not super familar with sticky

Mahmud-Khan: Edge/ie does not support sticky position. I was trying to apply CSS specific to edge/ie and set the position fixed for them. Talked to Mike and he suggested to set position: fixed above sticky. Tested it and from my understanding, if sticky is not supported it falls back to fixed.

KevinJJackson:
your main content is scrollable even when there is no content, might need to change the min-height on that a bit.

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -57,11 +63,13 @@
 
   .help-button {
     margin: 15px;
+    position: static;
```
mhemesath: this shouldn't be needed.

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -57,11 +63,13 @@
...
     text-align: right;
 
     @media screen and (min-width: $terra-medium-breakpoint) {
       bottom: 15px;
       margin: 0;
+      padding: 0;
```
mhemesath: this shouldn't be needed if the component doesn't have padding outside of the media query

packages/terra-consumer-nav/src/components/modal/Modal.scss
```diff
@@ -5,13 +5,15 @@
     background: var(--terra-consumer--modal-background, #fff);
     border-radius: 5px;
     left: 20px;
+    max-height: calc(100% - 80px);
```
mhemesath: why is this needed?

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -34,6 +34,12 @@
     white-space: break-word;
   }
 
+  .main-content {
+    @media screen and (max-width: $terra-medium-breakpoint) {
+      min-height: calc(100vh - 120px);
```
mhemesath: can you add a comment why we are doing this

packages/terra-consumer-layout/src/Layout.scss
Mahmud-Khan: Will remove that. Used padding instead of margin before so had to set padding 0 for mobile, not needed now.

packages/terra-consumer-nav/src/components/modal/Modal.scss
Mahmud-Khan: Scroll bar does not appear if max-height is not set. The modal header(title and close button) height is 60px and space below modal content is 10px/20px for mobile/tablet. Subtracted the height of the header and space below content from screen height.

packages/terra-consumer-layout/src/Layout.scss
Mahmud-Khan: The help button height is 40px and it has 20px margin top-bottom. Subtracted 80px initially from 100vh to stop the help button sticking on the bottom of the screen. However, even for empty content scroll appears like below
layout_scroll
Added some extra space to stop scroll when content empty.

mhemesath: I mean in the code :)

packages/terra-consumer-nav/src/components/modal/Modal.scss
mhemesath: I guess I don't understand why a help button fix needs to modify modal styles?

Mahmud-Khan: Sorry, there was a modal content scrolling issue in Nav. I fixed it as part of the same PR. I can create a separate PR.

PR #41, opened 2017-09-06
Layout margin themeable

WhiteLight1107:
Added themeable margins


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -29,6 +29,7 @@
 
   .main-container {
     flex-grow: 1;
+    margin: var(--terra-consumer-layout-margin, 0);
```
mhemesath: I feel like technically this should be padding..... (space between container and its content)

alex-bezek:
can we have a test case showing this works and can be set properly?

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -29,6 +29,7 @@
 
   .main-container {
     flex-grow: 1;
+    padding: var(--terra-consumer-layout-padding, 10px);
```
alex-bezek: Joey and I discussed this. The issue was if he adds a new css file for a specific example that sets root variables, it affects every example since its all in 1 bundle. 
I thought it would be fine to add a default of 10 and users can always override them. 
Later on we may need to think of a good strategy for this.

PR #42, opened 2017-09-07
stop using terra buttons for components we are applying our own class…

alex-bezek:
… to and overriding all the terra button styles

### Summary
Currently, we are consuming terra-button for most of our buttons. In the case of the help button, profile button, and close nav modal button, we are styling them completely differently by adding our own classname and overriding all of the base styles. While consuming this from Dex during a production asset build, the terra-core styles win out. Both are just a single class selector specificity and i'm assuming it just depends on how the classes get added to the final bundle. 

Since we are overriding all the styles, its fine for us to just use the html button itself and style it ourselves.


packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ChangeLog
 =========
 
+# 0.1.0-BETA.7 - (September XX, 2017)
```
KevinJJackson: i think this should still be in beta.6

PR #43, opened 2017-09-07
add support for site-alert banner

mhemesath:
### Summary
Add support for a site alert banner n the navigation

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.jsx
+++ b/packages/terra-consumer-layout/src/Layout.jsx
@@ -8,6 +8,12 @@ import styles from './Layout.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+
+  /**
+  * Object of configuration for the site alert.
+  */
+  siteAlert: PropTypes.string,
+
   /**
   * Object of configuration for the side navigation and profile.
   */
@@ -40,15 +46,20 @@ class Layout extends React.Component {
   }
 
   render() {
-    const { nav, helpItems, intl, ...customProps } = this.props;
+    const { nav, helpItems, siteAlert, intl, ...customProps } = this.props;
     return (
       <div>
+        {
+            siteAlert
+              ? <div className={cx('site-alert')}>{ siteAlert }</div>
+              : null
+        }
         <div className={cx('skip-container')}>
           <a className={cx('skip-to-main-content')} href="#main-container" id="skip-maincontent-link">
             {intl.formatMessage({ id: 'Terra.Consumer.Layout.skipcontent' })}
           </a>
         </div>
-        <div className={cx('layout', customProps.className)} {...customProps}>
+        <div className={cx('layout', { 'site-alert-active': siteAlert }, customProps.className)} {...customProps}>
           <Nav
             {...nav}
             isMobileNavOpen={this.state.isMobileNavOpen}
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -14,6 +14,20 @@
 }
 
 :local {
+
+  .site-alert {
+    position: fixed;
+    background: var(--terra-consumer-site-alert-background, #000);
+    color: var(--terra-consumer-site-alert-color, #FFF);
+    bottom: 0;
+    width: 100%;
+    left: 0;
+    height: var(--terra-consumer-site-alert-height, 30px);
+    text-align: center;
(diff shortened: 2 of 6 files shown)


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -63,6 +77,10 @@
       margin: 0;
       position: fixed;
       right: 15px;
+
+      :global(.site-alert-active) & {
+        bottom: calc(var(--terra-consumer-site-alert-height, 30px) + 15px);
```
mhemesath: not super thrilled with this approach, but its the best solution I can think of right now

packages/terra-consumer-layout/src/Layout.jsx
```diff
@@ -40,15 +46,20 @@ class Layout extends React.Component {
...
   render() {
-    const { nav, helpItems, intl, ...customProps } = this.props;
+    const { nav, helpItems, siteAlert, intl, ...customProps } = this.props;
     return (
       <div>
+        {
+            siteAlert
+              ? <div className={cx('site-alert')}>{ siteAlert }</div>
+              : null
+        }
```
alex-bezek: nit: this spacing might cause a lint error

packages/terra-consumer-layout/src/Layout.jsx
```diff
@@ -8,6 +8,12 @@ import styles from './Layout.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+
+  /**
+  * Object of configuration for the site alert.
+  */
+  siteAlert: PropTypes.string,
```
alex-bezek: so if this is being passed into layout as an object, how do you forsee something like dex using this? like a redux store that some lower level component triggers to update and adds a siteAlert into the store, which then propogates to this and renders the alert?

We can worry about this later, but layout wraps everything, so changing a prop will cause the whole page to re-render. I think we thought through this and solved it in theme provider with a middle container around the children content that specifies shouldcomponentupdate

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -9,6 +9,13 @@
     position: fixed;
     width: var(--terra-consumer-profile-link-width, 300px);
 
+    @media screen and (min-width: $terra-medium-breakpoint) {
+      :global(.site-alert-active) & {
+        bottom: var(--terra-consumer-site-alert-height, 30px)
```
alex-bezek: missing a comma causing your build to fail
```
lerna ERR! execute src/components/user-profile/UserProfile.scss
lerna ERR! execute  14:61  ✖  Expected a trailing semicolon   declaration-block-trailing-semicolon
```

packages/terra-consumer-layout/src/Layout.jsx
mhemesath: Site alerts are different than content alerts. A site alert is like an announcement for a scheudled downtime... it will be configured in the CMS and not change dynamically via interacting with app. 

But you do bring up a good concern.. and I would like to talk more about it

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -9,6 +9,13 @@
     position: fixed;
     width: var(--terra-consumer-profile-link-width, 300px);
 
+    @media screen and (min-width: $terra-medium-breakpoint) {
```
zhongyn: since there are other places using media query for mobile with
```
max-width: $terra-medium-breakpoint
```
so it should be $terra-medium-breakpoint +1 to avoid the inconsistency at that breakpoint

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -14,6 +14,20 @@
...
+  .site-alert {
+    position: fixed;
+    background: var(--terra-consumer-site-alert-background, #000);
+    color: var(--terra-consumer-site-alert-color, #FFF);
+    bottom: 0;
+    width: 100%;
+    left: 0;
+    height: var(--terra-consumer-site-alert-height, 30px);
+    text-align: center;
+    line-height: var(--terra-consumer-site-alert-height, 30px);
```
alex-bezek: If the line height and height aren't the same it won't be vertically centered anymore right? Is there a good way to handle custom heights with always guaranteeing centered text?

mhemesath:
Closing this review. The design was updated to put alerts at top. We'll defer this until next quarter when site alerts are formally added

PR #44, opened 2017-09-07
Fix for Nav mobile scroll for Firefox

javalkar-varun:
### Summary
When the screen is made small both vertically and horizontally, and there are lots of links (like the patient portal example) 2 links at the bottom don't show

Fix for Nav mobile scroll in Firefox:
firefox_scroll

@cerner/terra


javalkar-varun:
Created new PR from all new changes from the master.

PR #46, opened 2017-09-07
Fix for Nav - mobile scroll for Firefox

javalkar-varun:
### Summary
When the screen is made small both vertically and horizontally, and there are lots of links (like the patient portal example) 2 links at the bottom don't show

Before:
before_firefox_scroll

After:
firefox_scroll

@cerner/terra #

### DIFF
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -51,9 +51,11 @@
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
       position: absolute;
+      
+      /*stylelint-disable function-url-quotes*/
       @-moz-document url-prefix() {
+        height: 80% !important;
         min-height: 93vh;
-        height:80% !important;
         overflow-y: auto;
       }
     }
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -42,21 +42,18 @@
     user-select: none;
     width: var(--terra-consumer-nav-width, 320px);
     z-index: 2;
-
     @media screen and (max-width: $terra-tiny-breakpoint) {
       width: 100vw;
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
+      height: calc(100% - 60px) !important;
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
       position: absolute;
-      
       /*stylelint-disable function-url-quotes*/
       @-moz-document url-prefix() {
-        height: 80% !important;
-        min-height: 93vh;
-        overflow-y: auto;
+        min-height: max-content;
       }
     }
   }
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -38,6 +38,7 @@
     overflow-y: auto;
     padding-bottom: 60px;
     padding-top: var(--terra-consumer-nav-padding-top, 25px);
+    position: relative;
     top: 0;
     user-select: none;
     width: var(--terra-consumer-nav-width, 320px);
@@ -49,7 +50,6 @@
     @media screen and (max-width: $terra-medium-breakpoint) {
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
-      position: relative;
     }
   }
 
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -47,14 +47,9 @@
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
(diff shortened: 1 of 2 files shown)


mhemesath:
A simpler fix is to just ensure there is a bottom padding of the nav container equal to the height of the profile menu.

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -42,15 +42,19 @@
     user-select: none;
     width: var(--terra-consumer-nav-width, 320px);
     z-index: 2;
-
     @media screen and (max-width: $terra-tiny-breakpoint) {
       width: 100vw;
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
+      height: calc(100% - 60px) !important;
```
mhemesath: why important and where is 60px coming from?

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -42,15 +42,19 @@
...
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
       position: absolute;
+      /*stylelint-disable function-url-quotes*/
+      @-moz-document url-prefix() {
+        min-height: max-content;
```
mhemesath: are you sure this is needed? Also according to MDN that property value isn't supported in firefox. https://developer.mozilla.org/en-US/docs/Web/CSS/min-height

packages/terra-consumer-nav/src/Nav.scss
mhemesath: If the 60px is the profile height, i'm confused why line 39 is not fixing this problem as that is in place to address this issue.

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -42,15 +42,19 @@
...
       width: 100vw;
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
+      height: calc(100% - 60px) !important;
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
       position: absolute;
+      /*stylelint-disable function-url-quotes*/
+      @-moz-document url-prefix() {
```
mhemesath: I don't think you need this at all... just leave `min-height: max-content;`

packages/terra-consumer-nav/src/Nav.scss
mhemesath: can this just be applied on line 35? Also shouldn't need important

zhongyn:
@javalkar-varun can you try the padding solution suggested by mike?

zhongyn:
@javalkar-varun I think the root cause is line 53 `position: absolute`. Switching it to `relative` fixed the scroll issue. Tested it in chrome/firefox/ie.

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -42,15 +42,14 @@
...
-
     @media screen and (max-width: $terra-tiny-breakpoint) {
       width: 100vw;
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
       overflow: auto;
       padding-top: var(--terra-consumer--mobile-nav-padding-top, 0);
-      position: absolute;
+      position: relative;
```
zhongyn: can you move it outside of the media query. It should apply to both desktop and mobile

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -42,15 +42,14 @@
     user-select: none;
     width: var(--terra-consumer-nav-width, 320px);
     z-index: 2;
-
     @media screen and (max-width: $terra-tiny-breakpoint) {
       width: 100vw;
     }
 
     @media screen and (max-width: $terra-medium-breakpoint) {
       overflow: auto;
```
zhongyn: i think this is no needed anymore

javalkar-varun: yes, removed.

packages/terra-consumer-nav/src/Nav.scss
javalkar-varun: yes, moved. Thanks for the input.

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -38,19 +38,18 @@
     overflow-y: auto;
     padding-bottom: 60px;
     padding-top: var(--terra-consumer-nav-padding-top, 25px);
+    position: relative;
```
mhemesath: interesting, you just needed to add relative positioning? Does this fix it in IE too?

javalkar-varun: Yes, win10 Edge and IE
image

javalkar-varun: yes, win10 Edge and IE. I think height: 100% which is normal position but when position is set to relative, nav behaves w.r.t to normal position.

PR #47, opened 2017-09-07
Close nav on select

KevinJJackson:
### Summary

Added functionality within Layout to close nav when user clicks outside of it or presses the escape key.
Added functionality that closes mobile nav when a nav-link is selected.

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.jsx
+++ b/packages/terra-consumer-layout/src/Layout.jsx
@@ -30,7 +30,31 @@ class Layout extends React.Component {
     this.state = {
       isMobileNavOpen: false,
     };
+
     this.toggleNav = this.toggleNav.bind(this);
+    this.setReference = this.setReference.bind(this);
+    this.handleClickOutside = this.handleClickOutside.bind(this);
+  }
+
+  // Add event listener for clicks when component mounts
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleClickOutside);
+  }
+
+  // Remove event listener before breakdown of react component
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleClickOutside);
+  }
+
+  // set reference node without document.getElementById or ReactDOM
+  setReference(node) {
+    this.referenceNode = node;
+  }
+
+  handleClickOutside(event) {
+    if (this.referenceNode && !this.referenceNode.contains(event.target) && this.state.isMobileNavOpen) {
+      this.toggleNav();
+    }
   }
 
   toggleNav() {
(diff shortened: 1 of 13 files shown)


mhemesath:
Overlay was actually created for this use case.. I think we should just mount an overlay and ensure it is stacked behind the nav. Then if its clicked, close the nav and remove the overlay.

http://terra-ui.herokuapp.com/components/core/overlay

packages/terra-consumer-layout/src/Layout.jsx
```diff
@@ -55,6 +67,7 @@ class Layout extends React.Component {
             onRequestClose={this.toggleNav}
           />
           <main id="main-container" className={cx('main-container', this.state.isMobileNavOpen && 'nav-open')}>
+            <ResponsiveElement defaultElement={overlay} responsiveTo="window" medium={<div />} />
```
mhemesath: do we need responsive element? You could probably just always render the overlay

packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -129,7 +129,7 @@ class Nav extends React.Component {
         >
           <Button icon={<IconClose />} className={cx('close-button')} onClick={() => { onRequestClose(); }} />
           <NavLogo {...logo} />
-          <NavItems navItems={navItems} />
+          <NavItems navItems={navItems} handleClick={onRequestClose} />
```
mhemesath: will this cause problems if we ever have nav's open modals and thigns? If we won't do this today, then we can probably punt that problem

packages/terra-consumer-layout/src/Layout.jsx
KevinJJackson: we do. if we always render the overlay and the navOpen flag is set to true (for example resizing from mobile to desktop while nav is open), it breaks desktop by allowing scroll and forces the user to click main content once to close the nav so they can interact with the content itself.
image

packages/terra-consumer-nav/src/Nav.jsx
KevinJJackson: it shouldn't be a big issue, i imagine the user would click the link which opens the modal but would also silently close the nav behind the open modal.

packages/terra-consumer-layout/src/Layout.jsx
KevinJJackson: i tried setting css properties of visibility: hidden and display: none to the overlay & tried setting the same properties on a wrapper around the overlay but neither affected it.

mhemesath: ok

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -29,6 +29,7 @@
 
   .main-container {
     flex-grow: 1;
+    min-height: 100vh;
```
zhongyn: what does this one fix?

KevinJJackson: without a min height on the container, the overlay wouldn't span the rest of the screen. so if there was very little content on the screen, the user would have to click inside of the content to close the nav instead of anywhere on the screen

KevinJJackson: now that i think about it, 100% might be better than 100vh

KevinJJackson: added the light background on the overlay for visibility on the screenshots:
with min-height: 100%
image

without min-height:
image

PR #51, opened 2017-09-12
fixes so we arent passing classes to the terra button which get overr…

alex-bezek:
…iden

### Summary
When passing class names to terra button, we are using them to override values like display hidden based on media queries. This works sometimes, but since they have the same specificity, if the terra-button css comes after it in our final bundle it wins and the buttons always show. To fix this display based css is moved to a container around the button, whilst the couple required css attributes get !important per @mhemesath 's recommendation

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.scss
+++ b/packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.scss
@@ -6,6 +6,7 @@
     border: var(--terra-consumer-burger-border, 0);
     box-shadow: var(--terra-consumer-burger-box-shadow, 0);
     color: var(--terra-consumer-burger-color, #fff);
+    font-size: 20px;
 
     &:hover,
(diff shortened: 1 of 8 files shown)


packages/terra-consumer-nav/src/components/nav-burger-button/NavBurgerButton.scss
```diff
@@ -6,6 +6,7 @@
     border: var(--terra-consumer-burger-border, 0);
     box-shadow: var(--terra-consumer-burger-box-shadow, 0);
     color: var(--terra-consumer-burger-color, #fff);
+    font-size: 20px;
```
KevinJJackson: should this be theme-able?

alex-bezek: we can always add themable options later. I would rather keep them to a minimum unless we know we need it

PR #52, opened 2017-09-12
Cleanup custom properties

javalkar-varun:
### Summary
There are multiple places where custom properties are not unique. They should be separate and themed within Layout.scss within the :root clause: https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-layout/src/Layout.scss#L3

Naming convention followed to create custom property:
--terra-consumer-classname-propertyname

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -4,14 +4,14 @@
   --terra-consumer-body-background: url('https://wallpapercave.com/wp/5ynt7Gb.jpg');
   --terra-consumer-nav-active-item-background: var(--terra-consumer-body-background);
   --terra-consumer-mobile-panel-background: var(--terra-consumer-body-background);
-  --terra-consumer-body-layout-background: var(--terra-consumer-body-background);
+  --terra-consumer-layout-background: var(--terra-consumer-body-background);
   --terra-consumer-nav-item-color: #fff;
   --terra-consumer-toggle-header-hover-color: var(--terra-consumer-nav-item-color);
 
-  --terra-consumer-nav-width: 320px;
+  --terra-consumer-mobile-panel-width: 320px;
   --terra-consumer-nav-text-color: var(--terra-consumer-nav-item-color);
 
-  --terra-consumer-profile-link-width: var(--terra-consumer-nav-width);
+  --terra-consumer-profile-link-width: var(--terra-consumer-mobile-panel-width);
   --terra-consumer-profile-link-background: var(--terra-consumer-body-background);
   --terra-consumer-profile-link-color: var(--terra-consumer-nav-item-color);
   --terra-consumer-nav-item-hover-color: var(--terra-consumer-nav-item-color);
@@ -21,7 +21,7 @@
 :local {
   .layout {
     backface-visibility: hidden;
-    background: var(--terra-consumer-body-layout-background);
+    background: var(--terra-consumer-layout-background);
     background-attachment: fixed;
     background-size: cover;
     display: flex;
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -3,7 +3,10 @@
 :root {
   --terra-consumer-body-background: url('https://wallpapercave.com/wp/5ynt7Gb.jpg');
   --terra-consumer-nav-active-item-background: var(--terra-consumer-body-background);
+  --terra-consumer-mobile-panel-background: var(--terra-consumer-body-background);
+  --terra-consumer-body-layout-background: var(--terra-consumer-body-background);
   --terra-consumer-nav-item-color: #fff;
+  --terra-consumer-toggle-header-hover-color: var(--terra-consumer-nav-item-color);
 
   --terra-consumer-nav-width: 320px;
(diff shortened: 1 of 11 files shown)


packages/terra-consumer-nav/src/components/nav-help/NavHelpPopup.scss
```diff
@@ -11,7 +11,7 @@
   .close-button {
     background: none;
     border: 0;
-    color: var(--terra-consumer--popup-close-button-color, #fff);
+    color: var(--terra-consumer-close-button-color, #fff);
```
Mahmud-Khan: This style is specific to the close button in popup header. If we use --terra-consumer-close-button-color then it seems all the close buttons would use the same color. Should we follow the similar pattern used above in Modal.scss  **--terra-consumer-close-button-color** ?

javalkar-varun: Yes, since this is specific to Popup I will change it to `--terra-consumer-popup-close-button-color`. Similarly, hover and focus will change to `--terra-consumer-popup-close-button-hover-color` and `--terra-consumer-popup-close-button-focus-color` respectively.

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -3,21 +3,25 @@
 :root {
   --terra-consumer-body-background: url('https://wallpapercave.com/wp/5ynt7Gb.jpg');
```
bjankord: We should host this image on the HealtheLife S3 bucket and reference from there instead of wallpapercave.com

packages/terra-consumer-nav/src/components/nav-items/NavItem.scss
```diff
@@ -2,7 +2,7 @@
 
 :local {
   .item {
-    color: var(--terra-consumer-nav-item-color, #fff);
+    color: var(--terra-consumer-item-color, #fff);
```
bjankord: Should the variables in this file keep the `nav` word in the variable name? They all look specific to nav.

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.scss
```diff
@@ -13,7 +13,7 @@
   }
 
   .img {
-    color: var(--terra-consumer-logo-text-color, #3d3d3d);
+    color: var(--terra-consumer-img-color, #3d3d3d);
```
bjankord: nit: I'd recommend `--terra-consumer-logo-img-color` to keep it specific to logo instead of general to img.

packages/terra-consumer-layout/src/Layout.scss
javalkar-varun: @mhemesath @alex-bezek any thoughts on this?

mhemesath: that isn't even the image we want I think...

mhemesath: Varun, I think I'm ok with adding the image to S3...

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -3,21 +3,25 @@
...
   --terra-consumer-nav-active-item-background: var(--terra-consumer-body-background);
+  --terra-consumer-mobile-panel-background: var(--terra-consumer-body-background);
+  --terra-consumer-layout-background: var(--terra-consumer-body-background);
   --terra-consumer-nav-item-color: #fff;
+  --terra-consumer-toggle-header-hover-color: var(--terra-consumer-nav-item-color);
```
alex-bezek: nit: use the word toggler instead of toggle. Toggle makes me think its a boolean or something dictating if it should toggle

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -3,21 +3,25 @@
...
   --terra-consumer-nav-text-color: var(--terra-consumer-nav-item-color);
 
-  --terra-consumer-profile-link-width: var(--terra-consumer-nav-width);
+  --terra-consumer-profile-link-width: var(--terra-consumer-mobile-panel-width);
   --terra-consumer-profile-link-background: var(--terra-consumer-body-background);
   --terra-consumer-profile-link-color: var(--terra-consumer-nav-item-color);
-  --terra-consumer-logo-text-color: var(--terra-consumer-nav-item-color);
+  --terra-consumer-nav-item-hover-color: var(--terra-consumer-nav-item-color);
+  --terra-consumer-profile-background: var(--terra-consumer-body-background);
 }
```
alex-bezek: do all these variables belong in layout? Alot of them are only used in nav specific component

mhemesath: +1 We need to remove all the variables from layout and just have defaults set everywhere. If we want to sync their default value with sass variables, I'm OK with that.

mhemesath:
So, this branch is not in sync with master. I recommend we abandon it and just create a fresh branch from master and redo this. Its seriously not worth trying to resolve the conflicts

javalkar-varun:
Close this PR and open new one. There are too many conflicts. It is not in sync with the master.

PR #54, opened 2017-09-12
Add sign out to profile

KevinJJackson:
### Summary
Move logic for determining which link to render to profile component.
Add logic to render sign out as the main button when certain criteria are met.

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
+++ b/packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import Arrange from 'terra-arrange';
 import classNames from 'classnames/bind';
 import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
+import Button from 'terra-button';
 import IconEllipses from 'terra-icon/lib/icon/IconEllipses';
 import IconPerson from 'terra-icon/lib/icon/IconPerson';
 import ProfileLinks from './ProfileLinks';
@@ -31,10 +32,8 @@ const propTypes = {
    * The path to the login page.
    */
   signinUrl: PropTypes.string,
-  /**
-   * Determiniate of whether profile should render as signin link or profile button.
-   */
-  isSignIn: PropTypes.bool,
+  isExternal: PropTypes.bool,
+  target: PropTypes.string,
   /**
    * The content of the each profile items.
    */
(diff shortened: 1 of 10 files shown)


packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -50,26 +47,30 @@ const propTypes = {
...
-          fitStart={<div className={cx('avatar')}>{avatar}</div>}
-          fill={<FormattedMessage id="Terra.Consumer.UserProfile.signin" />}
-          align="center"
-        />
-      </button>
-    );
+  const avatarIcon = <div className={cx('avatar')}>{avatar || <IconPerson />}</div>;
+
+  const setStaticProfile = (url, translationId) => (
+    <Button className={cx('popup-button')} href={url}>
```
alex-bezek: don't pass the class to the terra button. It causes bugs (see here https://github.com/cerner/terra-consumer/pull/51 ). do you need terra button since we are overriding all of its look?

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -50,26 +47,30 @@ const propTypes = {
...
+      <Arrange
+        fitStart={avatarIcon}
+        fill={<FormattedMessage id={translationId} />}
+        align="center"
+      />
+    </Button>
+  );
+
+  if (signinUrl && !(avatar || userName)) {
```
alex-bezek: nit: add a comment describing what this logic does and why we want it this way

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -50,26 +47,30 @@ const propTypes = {
...
-        <Arrange
-          fitStart={<div className={cx('avatar')}>{avatar}</div>}
-          fill={<FormattedMessage id="Terra.Consumer.UserProfile.signin" />}
-          align="center"
-        />
-      </button>
-    );
+  const avatarIcon = <div className={cx('avatar')}>{avatar || <IconPerson />}</div>;
+
+  const setStaticProfile = (url, translationId) => (
```
alex-bezek: what are we setting? aren't we just returning a profile arrange?

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: i tried just using an a tag here, but the styles of the 'popup-button' were super glitchy. I could use a regular button and have a function to set the window.location but that seems like a last resort. Are there other options I could try here?

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: I just wrote this function so i didn't have to duplicate code for the sign-in/sign-out profile link, since all that changed between the two is the text and the link

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
```diff
@@ -32,10 +33,6 @@ const propTypes = {
    */
   signinUrl: PropTypes.string,
```
alex-bezek: can you update the comments here to clearly reflect what might render/happen if certain things are passed in?
I'm not sure we should change whats rendering based off if this is passed in (maybe just no sign in link at all if its not passed in?). I think most apps will have a static signIn and signOut url so they will always be passing them in

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
alex-bezek: I'm just being picky about the name. Drying up the code is fine.

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: I'll try the same technique used in the mentioned pr

alex-bezek: can you try what i did in my other PR. put our class on a wrapping div that handles any position based things, remove alot of the background, border, etc stuff that using a button variant of link fixes. If there is an abosulte need for some override of terra button styles, @mhemesath said we could use important sparingly

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: yeah it could get confusing for people that didnt write the code, will comment this

packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx
KevinJJackson: will come up with a better name haha

mhemesath:
lets hold off on this for a bit.. I need to verify expected behavior

KevinJJackson:
this pr is no longer needed as the logic as changed and is switched to no longer render the profile if no object is passed. see #73

PR #56, opened 2017-09-12
removing brackets from language option

amf9t2:
removing brackets from language option for now as having issue in rendering language item with brakctes in rtl

One option we have to fix this by appending markup (RLM OR LRM) to values similar how [portal does](https://github.cerner.com/patient-portal/style-guide/blob/master/style_guide/utils/translations.py#L96)

 . Currently the data flow to component is dynamic and not sure if its recommended to check dir in react component  [profile link text](https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx#L59) to append unicode respectively. Believing it is not straight forward to append the markup I opted to remove the brackets for now.
image

### DIFF
--- a/packages/terra-consumer-layout/tests/nightwatch/PortalLayout.jsx
+++ b/packages/terra-consumer-layout/tests/nightwatch/PortalLayout.jsx
@@ -278,7 +278,7 @@ const data = {
           target: '_self',
           url: '#',
           isExternal: true,
-          text: 'English (United States)',
+          text: 'English - United States',
           subItems: [
             {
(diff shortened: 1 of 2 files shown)


amf9t2:
image

mhemesath:
I think this is because we're not setting the directional special characters in the text. I logged an internal issue about this

alex-bezek:
@amf9t2 can we close this?

mhemesath:
I think so

PR #58, opened 2017-09-13
fix the hamburger position and consume fix for pop up

alex-bezek:
### Summary
Hamburger position should maintain relative to its normal flow to keep within the main content and get its padding. But it should be bumped over slightly without affecting flow

@cerner/terra


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -123,10 +123,9 @@
   .nav-burger {
     display: block;
     font-size: 20px;
-    left: 15px;
-    padding: 0;
-    position: absolute;
-    top: 10px;
+    left: -13px;
```
mhemesath: where is 13px coming from?

packages/terra-consumer-nav/src/components/modal/Modal.scss
```diff
@@ -30,38 +30,37 @@
...
-    border: 0;
-    color: var(--terra-consumer--modal-close-button-color, #fff);
-    font-size: 16px;
+  .close-button-container {
     margin: 0;
     padding: 0;
     position: absolute;
     right: 20px;
-    top: 5px;
+    top: 0;
```
amf9t2: what's difference between not adding `top` and  adding `top: 0` ?

packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx
```diff
@@ -53,13 +52,13 @@ class ProfileLinks extends React.Component {
         const toggleIcon = isOpen ? <IconChevronUp /> : <IconChevronDown />;
         return (
           <div key={linkItem.text} className={cx(isOpen && 'open')}>
-            <Button onClick={() => { this.handleToggle(index); }} className={cx('link', 'toggler-wrapper')}>
+            <button onClick={() => { this.handleToggle(index); }} className={cx('link', 'toggler-wrapper')}>
               <Arrange
                 align="stretch"
```
amf9t2: align `top` didn't fix the icon alignment  ?

packages/terra-consumer-nav/src/components/modal/Modal.scss
alex-bezek: removed it

PR #59, opened 2017-09-13
Navtext dangerously

Mahmud-Khan:
### Summary
Rendered translated texts using dangerouslySetInnerHTML.

Before:
text_render_before

After:
text_render_after

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/dangerous-html/DangerousHtml.jsx
+++ b/packages/terra-consumer-nav/src/components/dangerous-html/DangerousHtml.jsx
@@ -9,6 +9,7 @@ const DangerousHtml = ({
   text,
   ...customProps
 }) => (
+  // eslint-disable-next-line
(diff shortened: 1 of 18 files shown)


packages/terra-consumer-nav/src/components/dangerous-html/DangerousHtml.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  text: PropTypes.string.isRequired,
+};
+
+const DangerousHtml = ({
```
mhemesath: I Almost prefer this to be called SafeHTML... That follows closer to django and rails patterns. I get the whole call it dangerous so ppl don't use it thing.. but in this case this is HTML we deem safe to inject. FWIW, this could be a candidate for terra-core potentially? cc: @bjankord

PR #60, opened 2017-09-13
logo text fix - firefox

amf9t2:
skipping src attribute if url is not passed in.

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
@@ -11,7 +11,7 @@ const propTypes = {
    */
   url: PropTypes.string,
   /**
-   *  The alternate text that is read by screen readers or displayed if the image fails to load.
+   *  The alternate text that is read by screen readers are displayed if the image fails to load.
    */
   altText: PropTypes.string,
   /**
@@ -41,10 +41,14 @@ const NavLogo = ({
   link,
   ...customProps
 }) => {
-  const image = (link) ?
-    <SmartLink {...link}> <img className={cx('img')} src={url} alt={altText} /> </SmartLink> :
-    <img className={cx('img')} src={url} alt={altText} />;
-  const body = (isCard && !!url) ? <Card.Body> {image} </Card.Body> : image;
+  const imgProps = {
+    className: cx('img'),
+    ...url && { src: url },
+  };
+
+  const imgTag = (<img alt={altText} {...imgProps} />);
+  const imgContent = (link) ? <SmartLink {...link}>{imgTag} </SmartLink> : imgTag;
+  const body = (isCard && !!url) ? <Card.Body> {imgContent} </Card.Body> : imgContent;
   const domNode = (isCard && !!url) ? Card : 'div';
   const logoClassNames = cx(
     'logo-container',
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -22,6 +22,7 @@ ChangeLog
 - Padding on nav items changed to be align with profile
 - Rendered translated texts using component SafeHtml.
 - Signout and signin links are now 'a' tags to utilize the href attribute.
+- Fixed logo text issue in Firefox
 
 ------------------
 


amf9t2:
image

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -41,9 +42,15 @@ const NavLogo = ({
   link,
   ...customProps
 }) => {
+  const imgProps = {
+    className: cx('img'),
+  };
+  if (url) {
```
KevinJJackson: you can condense this into a neat little es6 operation:
```
const imgProps = {
  className: cx('img'),
  ...url && { src: url },
};
```

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -41,9 +42,15 @@ const NavLogo = ({
...
+    imgProps.src = url;
+  }
   const image = (link) ?
-    <SmartLink {...link}> <img className={cx('img')} src={url} alt={altText} /> </SmartLink> :
-    <img className={cx('img')} src={url} alt={altText} />;
+    <SmartLink {...link}> <img alt={altText} {...imgProps} /> </SmartLink> :
```
amf9t2: lint error forced me to keep the alt attribute inside the image tag.

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -41,9 +42,15 @@ const NavLogo = ({
...
+    <img alt={altText} {...imgProps} />;
```
KevinJJackson: np: perhaps split this out into a var instead of having it written out twice

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
amf9t2: [updated](https://github.com/cerner/terra-consumer/pull/60/files)

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
amf9t2: make sense [updated](https://github.com/cerner/terra-consumer/pull/60/files)

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -30,6 +30,7 @@ const propTypes = {
 const defaultProps = {
   altText: 'Client Logo',
   isCard: false,
+  url: undefined,
```
alex-bezek: this isn't needed is it? should be undefined anyway

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -41,10 +42,14 @@ const NavLogo = ({
   link,
   ...customProps
 }) => {
-  const image = (link) ?
-    <SmartLink {...link}> <img className={cx('img')} src={url} alt={altText} /> </SmartLink> :
-    <img className={cx('img')} src={url} alt={altText} />;
-  const body = (isCard && !!url) ? <Card.Body> {image} </Card.Body> : image;
+  const imgProps = {
+    className: cx('img'),
+    ...url && { src: url },
```
alex-bezek: I don't understand the `...url`. The spread operator is for objects, but url is a string

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -41,10 +42,14 @@ const NavLogo = ({
...
+  };
+
+  const imgTag = (<img alt={altText} {...imgProps} />);
+  const imgContent = (link) ? <SmartLink {...link}>{imgTag} </SmartLink> : imgTag;
+  const body = (isCard && !!url) ? <Card.Body> {imgContent} </Card.Body> : imgContent;
```
alex-bezek: is the `!!url` double negative some crazy javascript truthy hack?

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
KevinJJackson: the `...` is for the object. url is just the bool check where if it returns false, it appends nothing to the object. returning true, spreads the return object on the right side of the equation `{src: url}` onto the parent object (imgProps)

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
KevinJJackson: as @WhiteLight1107 stated above, this is a truthify statement. any value that is not: false, 0, "", null, undefined, or NaN, will return true, otherwise false. https://developer.mozilla.org/en-US/docs/Glossary/Truthy

WhiteLight1107: I love the name "Thruthy"

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
KevinJJackson: it's also worth noting that it is able to be written this way since in the order of operations, `&&` has a higher precedence than `...`

zhongyn: if url is falsey, what does it return?

amf9t2: @zhongyn url is falsey you mean invalid url or empty ? 
if empty then `...url` will be false and src will not be added in imgProps. 
If invalid url then it will still be added but browser will render alt text in this case.

KevinJJackson: @zhongyn if url is falsey, `...` essentially doesn't do anything, since it would be trying to spread a boolean 'false' which spreads nothing.

wrote a quick codepen if you want to play around with it: https://codepen.io/anon/pen/veEPpK?editors=1012

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
amf9t2: yup by default props will be undefined. [updated](https://github.com/cerner/terra-consumer/pull/60/commits/a79dc69228e7473e8638480a670e14cd48dfb274)

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
KevinJJackson: there was also a question raised here: https://github.com/tc39/proposal-object-rest-spread/issues/45 in which it's stated that `...` mirrors `Object.assign` and
> Object.assign will treat all primitives (like undefined) as the same as {}

which therefore spreads nothing

alex-bezek: thats neat. seems tricky but i'm ok with it.

PR #66, opened 2017-09-15
Fixed modal close button alignment and padding issue

Mahmud-Khan:
### Summary
- Fixed modal close button alignment issue.
- Changed padding to match UI callout.

Before:
before

After:

after

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/modal/Modal.scss
+++ b/packages/terra-consumer-nav/src/components/modal/Modal.scss
@@ -53,9 +53,8 @@
   .close-button {
     color: var(--terra-consumer--modal-close-button-color, #fff) !important;
     font-size: 16px !important;
-    margin: 0;
-    padding: 0;
-
+    margin: 0 !important;
+    padding: 0 !important;
 
     &:hover {
       color: var(--terra-consumer--modal-close-button-hover-color, #777) !important;
--- a/packages/terra-consumer-nav/src/components/modal/Modal.scss
+++ b/packages/terra-consumer-nav/src/components/modal/Modal.scss
@@ -43,6 +43,7 @@
     padding: 0;
     position: absolute;
     right: 20px;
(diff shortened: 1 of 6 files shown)


packages/terra-consumer-nav/src/components/modal/Modal.scss
```diff
@@ -52,6 +53,9 @@
   .close-button {
     color: var(--terra-consumer--modal-close-button-color, #fff) !important;
     font-size: 16px !important;
+    margin: 0;
+    padding: 0;
```
zhongyn: i seems @alex-bezek is working on the same fix for dex?

mhemesath: @alex-bezek  we probably need !important here don't we?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -42,8 +42,8 @@ class NavHelpContent extends React.Component {
           <Arrange
             className={cx('help-item-text')}
             align="stretch"
-            fitStart={<div>{content.icon}</div>}
-            fill={<div className={cx('item-text-padding')}><SafeHtml text={content.text} /></div>}
+            fitStart={content.icon && <div className={cx('icon-text-padding')}>{content.icon}</div>}
```
mhemesath: trick.. if there isn't an icon we can do this in css
```
.icon-text-padding:empty {
   display: none;
}
```
This lets us keep our JSX a little cleaner

packages/terra-consumer-nav/src/components/modal/Modal.scss
alex-bezek: yes. anything being passed to a terra component via a class needs important

PR #68, opened 2017-09-18
get rid of all the css root variables

alex-bezek:
### Summary
variables declared in the root of our css cause issues in the consuming application. if they are present in the app bundle, and the app provides its root, and they end up in the same bundle, we are at the mercy of webpack to put the app's root below this repo's root.

Instead of relying on this, we should not use root in our module's css

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -1,31 +1,10 @@
 @import './variables';
 
-:root {
-  // Body
-  --terra-consumer-body-background-image: url('../background.jpg');
-  --terra-consumer-body-background-color: #c7d4ea;
-
-  // Nav
-  --terra-consumer-nav-background-color: rgba(255, 255, 255, 0.25);
-  --terra-consumer-nav-width: 320px;
-  --terra-consumer-nav-link-color: #3d3d3d;
-  --terra-consumer-nav-active-link-background: var(--terra-consumer-body-background-image);
-  --terra-consumer-nav-text-color: var(--terra-consumer-nav-link-color);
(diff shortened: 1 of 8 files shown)


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -1,31 +1,10 @@
...
 
-:root {
-  // Body
-  --terra-consumer-body-background-image: url('../background.jpg');
-  --terra-consumer-body-background-color: #c7d4ea;
-
-  // Nav
-  --terra-consumer-nav-background-color: rgba(255, 255, 255, 0.25);
-  --terra-consumer-nav-width: 320px;
-  --terra-consumer-nav-link-color: #3d3d3d;
```
KevinJJackson: can we default all of the nav text color to this instead of the white that's currently there. I think its just nav text, profile text and possibly the hamburger icon as well

alex-bezek: updated

KevinJJackson: looks like two more places:
https://github.com/cerner/terra-consumer/blob/a4c131631806907b672f188f1f3f79ddc5abad53/packages/terra-consumer-nav/src/components/nav-toggler/NavToggler.scss#L7
https://github.com/cerner/terra-consumer/blob/a4c131631806907b672f188f1f3f79ddc5abad53/packages/terra-consumer-nav/src/Nav.scss#L49

also, badge background should remain white:
https://github.com/cerner/terra-consumer/blob/a4c131631806907b672f188f1f3f79ddc5abad53/packages/terra-consumer-nav/src/components/nav-items/NavItem.scss#L44
image

PR #71, opened 2017-09-18
Allow the bar above the profile in mobile to be configured independen…

alex-bezek:
…tly of the background color

### Summary
Some clients don't want this set, some want it to be any color they want, etc

### Additional Details
To keep this change passive for you, just set the new variable equal to your old existing `--terra-consumer-body-background-color`

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
+++ b/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
@@ -10,7 +10,7 @@
       background: none;
       background-attachment: none;
       background-size: none;
-      border-top: 3px solid var(--terra-consumer-body-background-color, #c7d4ea);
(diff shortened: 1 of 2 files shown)


packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -10,7 +10,7 @@
       background: none;
       background-attachment: none;
       background-size: none;
-      border-top: 3px solid var(--terra-consumer-body-background-color, #c7d4ea);
+      border-top: 3px solid var(--terra-consumer-profile-link-border-top, #c7d4ea);
```
mhemesath: this is limited to the border-color.... Either change the name to border color or allow it to modify the whole border. e.g.
```csss
border-top: var(--terra-consumer-profile-link-border-top, 3px solid #c7d3ea);
```

PR #75, opened 2017-09-21
fix react router with full domain name links

alex-bezek:
### Summary
For react router internal links, strip the domain/host off the url before providing it to the link to prevent linking to something like http://localhost:8080/http:localhost:8080/my-page

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/components/smart-link/SmartLink.jsx
+++ b/packages/terra-consumer-nav/src/components/smart-link/SmartLink.jsx
@@ -25,19 +25,30 @@ const SmartLink = ({
   handleClick,
   children,
   ...customProps
-}) => (
-  (isExternal || target === '_blank') ?
-    <a {...customProps} onClick={handleClick} target={target} href={url}>{children}</a> :
+}) => {
+  if (isExternal || target === '_blank') {
+    return (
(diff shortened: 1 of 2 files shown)


packages/terra-consumer-nav/src/components/smart-link/SmartLink.jsx
```diff
@@ -25,19 +25,30 @@ const SmartLink = ({
...
-  (isExternal || target === '_blank') ?
-    <a {...customProps} onClick={handleClick} target={target} href={url}>{children}</a> :
+}) => {
+  if (isExternal || target === '_blank') {
+    return (
+      <a {...customProps} onClick={handleClick} target={target} href={url}>{children}</a>
+    );
+  }
+  const a = document.createElement('a');
+  a.href = url;
```
amf9t2: can we set href while creating  element ?

`document.createElement('a', {href: url}); `

alex-bezek: that is a react create element api. It doesn't seem to work for document create element
```
const x = document.createElement('a', {href: 'test'});

undefined
x
<a>​</a>​
```

kafkahw:
+1

PR #76, opened 2017-09-21
Ipad landscape background fix and nav burger button fix at 992 pixel

amf9t2:
We want either mobile view  or plain background same as mobile for ipad landscape scenario.

I tried multiple option for changing to mobile view but those are feasible because it require breakpoint change to large breakpoint in all cases.

so opted to make the background same as mobile with media query fix using $terra-large-breakpoint.


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -102,15 +103,15 @@
   }
 
   .nav-burger {
-    display: block;
```
amf9t2: this change fix the burger button issue. It wasn't displaying at 992px screen. Figured out we were referring two different ways min-width and max-width to switch styles.

amf9t2:
ipad_landscape

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -8,7 +8,8 @@
     background-size: cover;
     overflow-x: hidden;
 
-    @media screen and (max-width: $terra-medium-breakpoint) {
+    @media screen and (max-width: $terra-medium-breakpoint),
+      screen and (orientation: landscape) and (max-width: $terra-large-breakpoint) {
```
mhemesath: it looks like desktop computers are landscaped orientation (I assumed they didn't have orientation). This means this isn't an effective target for mobile...

mhemesath: I searched and the only way I can determine to target mobile browsers (the ones that don't support this) is by doing
```css
@supports (-webkit-overflow-scrolling: touch) {
  background-image: none;
}
```

This is actually effective enough that you can remove the medium breakpoint media query to apply it.

amf9t2: using orientation with combination of max-width will limit this change for mobile devices only (except ipad pro).  `@supports (-webkit-overflow-scrolling: touch)` approach also works -a   alternate way to handle this and can only be tested on actual devices. So we decided to keep the combination of orientation and max width.

PR #77, opened 2017-09-21
ie nav click hijacked by background

zhongyn:
### BEFORE
nav-click-before
### AFTER
nav-click-after

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -28,7 +28,7 @@
     min-width: var(--terra-consumer-nav-width, 320px);
     position: relative;
     width: var(--terra-consumer-nav-width, 320px);
-    z-index: 100;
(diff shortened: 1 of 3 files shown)


packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -28,7 +28,7 @@
     min-width: var(--terra-consumer-nav-width, 320px);
     position: relative;
     width: var(--terra-consumer-nav-width, 320px);
-    z-index: 100;
+    z-index: 200;
```
mhemesath: This works because its likely you were actually clicking the overlay which had a matching [z-index of 100](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.scss#L15).  IMO, overlay should have a higher z-index unless @bjankord has a reason for it being at 100. For now this fix is OK IMO

PR #83, opened 2017-10-04
adding fix for profile links close

alex-bezek:
### Summary
clicking on a link inside a modal wasn't triggering a close of the modal. This wasnt discovered with portal since it refreshes the page, but with dex using react router, it would navigate the user behind the modal but not show them. Now any clicks close the modal and nav panel

[CONTRIBUTORS.md] file signifies agreement to all rights and reservations provided by the [License].

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.scss
+++ b/packages/terra-consumer-layout/src/Layout.scss
@@ -31,7 +31,7 @@
     width: var(--terra-consumer-nav-width, 320px);
     z-index: 200;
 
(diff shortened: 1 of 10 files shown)


mhemesath:
did you verify this won't close the nav if a toggler is clicked?

alex-bezek:
@mhemesath yes the toggler doesn't close it. you can test it here
https://terra-consumer-pr-83.herokuapp.com/static/#/tests/layout-tests/portal
This is because the click handler is passed all the way to the SmartLink, and the toggler isn't a SmartLink

PR #84, opened 2017-10-05
badge accepts strings

KevinJJackson:
### Summary
Change to badges to allow passing of strings into the badge, instead of just numbers. This allows for users to customize what displays such as (  !  ) or (  1  ).

### Additional Details
Putting text through SafeHtml so users can put translated text or symbols as they desire into the badge.
Empty strings should not be rendered as a badge. taken care of by use of .trim() and truthy check.

@mhemesath @alex-bezek @zhongyn

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItem.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItem.jsx
@@ -54,13 +54,16 @@ const NavItem = ({
   ...customProps
 }) => {
   const activeClass = cx('active');
+  const badgeString = `${badgeValue}`;
 
-  const itemLabel = (<div>
-    <SafeHtml text={text} />
-    { badgeValue > 0 &&
-      <div className={cx('badge')}>{badgeValue}</div>
-    }
-  </div>);
+  const itemLabel = (
+    <div>
+      <SafeHtml text={text} />
+      { (!!badgeValue && badgeString.trim()) &&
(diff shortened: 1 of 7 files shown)


packages/terra-consumer-nav/src/components/nav-items/NavItem.jsx
```diff
@@ -54,20 +54,24 @@ const NavItem = ({
...
-  const itemLabel = (<div>
-    <SafeHtml text={text} />
-    { badgeValue > 0 &&
-      <div className={cx('badge')}>{badgeValue}</div>
-    }
-  </div>);
+  const itemLabel = (
+    <div>
+      <SafeHtml text={text} />
+      { (!!badgeValue && badgeString.trim()) &&
```
mhemesath: won't this fail on `0`?

KevinJJackson: yup, that's the point though, unless we want to display badge that has 0. That, to me, seems counter-productive.

mhemesath: yeah we want to display whatever value is passed unless its null or empty string.

KevinJJackson: okay will update this

KevinJJackson:
test scenarios:
image

PR #86, opened 2017-10-09
Set default punch-through color to transparent

mhemesath:
Fixes the punch-thru effect by making the border transparent by defaul.t

### DIFF
--- a/packages/terra-consumer-nav/src/Nav.scss
+++ b/packages/terra-consumer-nav/src/Nav.scss
@@ -24,7 +24,7 @@
     position: fixed;
     width: var(--terra-consumer-profile-link-width, 320px);
 
-    @supports (-webkit-overflow-scrolling: touch) {
+    @media screen and (max-width: $terra-medium-breakpoint) {
       position: static;
     }
   }
--- a/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
+++ b/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
@@ -98,11 +98,6 @@
     text-align: left;
     width: 100%;
 
-    @supports (-webkit-overflow-scrolling: touch) {
-      background: inherit;
-      margin-top: 0;
-    }
-
     &:hover,
     &:focus {
       background: var(--terra-consumer-profile-hover-background, rgba(255, 255, 255, 0.4));
--- a/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
+++ b/packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
@@ -7,9 +7,10 @@
     background-size: cover;
     margin-top: 5px;
 
-    @media screen and (max-width: $terra-medium-breakpoint) {
+    @supports (-webkit-overflow-scrolling: touch) {
       background-image: none;
     }
+
     @media screen and (max-width: $terra-tiny-breakpoint) {
       width: 100vw;
     }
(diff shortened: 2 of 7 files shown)


packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -97,7 +97,6 @@
     cursor: pointer;
     display: block;
     font-size: var(--terra-consumer-profile-link-font-size, 14px);
-    margin-top: 3px;
```
zhongyn: it seems this removes the punch-through in mobile

alex-bezek: i tested setting 
```
--terra-consumer-profile-link-border-top: 3px solid #9BDAF1
```
which is what we have in dex. and it shows up fine
image

mhemesath: If the border isnt an explicit color it removes it yes. I can try to find a fix to that

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -2,17 +2,14 @@
 
 :local {
   .profile {
-    background: var(--terra-consumer-profile-link-border-background, url('http://d320j1w59ws5w0.cloudfront.net/terra-consumer-background-images.jpg'));
+    background: var(--terra-consumer-profile-link-border-background, #c7d4ea url('http://d320j1w59ws5w0.cloudfront.net/terra-consumer-background-images.jpg'));
     background-attachment: fixed;
     background-size: cover;
-    border-top: var(--terra-consumer-profile-link-border-top, 3px solid #c7d3ea);
     margin-top: 5px;
```
mhemesath: i actually don't think we need this nor do  i think it does anything..

zhongyn:
+1

packages/terra-consumer-nav/src/Nav.scss
```diff
@@ -24,7 +24,7 @@
     position: fixed;
     width: var(--terra-consumer-profile-link-width, 320px);
 
-    @supports (-webkit-overflow-scrolling: touch) {
+    @media screen and (max-width: $terra-medium-breakpoint) {
```
mhemesath: this change is complicated, basically we always want the profile bar static on mobile menu...

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
```diff
@@ -2,16 +2,13 @@
...
 
-    @supports (-webkit-overflow-scrolling: touch) {
-      background: none;
-      background-attachment: none;
-      background-size: none;
+    @media screen and (max-width: $terra-medium-breakpoint) {
```
mhemesath: small screen sizes never has a fixed background behind the menu, so the profile menu never should on small screens either

mhemesath: @amf9t2 ^^

packages/terra-consumer-nav/src/Nav.scss
mhemesath: @amf9t2 ^^

packages/terra-consumer-nav/src/components/user-profile/UserProfile.scss
amf9t2: couple points I see after this change
1. Ipad landscape view is same as desktop. 
2. Mobile view has different background in content area and navigation except ipad landscape.

amf9t2: discussed with @mhemesath due to issue in links because of background image we will continue with this change.

KevinJJackson:
should hover effect hide the punch through?

base:
image

hover:
image

mhemesath:
@zhongyn  are we good to merge this?

zhongyn:
yes

PR #87, opened 2017-10-09
early mobile view fix 

amf9t2:
cause - using large breakpoint

fix - using  @supports (-webkit-overflow-scrolling: touch)

One can only test in actual devices or simulator.


amf9t2:
ipad_landscape_simultor

PR #91, opened 2017-10-27
Make font-family a custom property

gauthamsriman:
### Summary
Make font-family configurable in DEX's layout.

JIRA: https://jira2.cerner.com/browse/PORTALDEV-39697


alex-bezek:
Code change looks good. 
The broken build is being fixed as part of a different pull request that should be merged soon.
I would recomend throwing some custom font in here
https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-layout/tests/nightwatch/DexH%40CTheme.scss
using that variable you just defined and verify it works as expected

packages/terra-consumer-layout/tests/nightwatch/DexH@CTheme.scss
```diff
@@ -15,4 +15,6 @@
   --terra-consumer-body-background-color: rgb(155, 218, 241);
   --terra-consumer-nav-background-color: rgba(255, 255, 255, 0.3);
   --terra-consumer-nav-active-link-background: #9BDAF1 url('https://d320j1w59ws5w0.cloudfront.net/3Q6UT7T_clinic-background.jpg');
+
+  --terra-consumer-font-family: 'Roboto', sans-serif;
```
alex-bezek: you should set it to something else, and have your solution designer or ux designer verify things are working correctly. This here doesn't really do much since you are just setting it back to the default

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -15,7 +15,7 @@
 
   .layout {
     display: flex;
-    font-family: 'Roboto', sans-serif;
+    font-family: var(--terra-consumer-font-family, 'Roboto', sans-serif);
```
mhemesath: terra base actually has this property. We should remove it from layout and just use there:

https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.scss#L8

alex-bezek: @mhemesath, should this jira be used to also integrate with terra base? technically its listed as a dependency, we aren't using it though.
Should that be something done in layout? or the consuming repo (like healthelastic-react)?

cc @bjankord 
where is the terra core layout code? i can't seem to find it. curious what they are doing.

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -15,7 +15,7 @@
 
   .layout {
     display: flex;
-    font-family: var(--terra-consumer-font-family, 'Roboto', sans-serif);
+    font-family: var(--terra-base-font-family, var(--terra-consumer-font-family, 'Roboto', sans-serif));
```
mhemesath: this won't work in legacy browsers... just make it roboto and remove the terra-consumer-font-family altogether.

bjankord: Another option would be remove font-family here and import `baseStyles.js` similar to how we do in terra-core components: https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L4

You'll also want to [update your version of terra-base](https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-layout/package.json#L25-L42) to `v2.x`

mhemesath: Yeah I had suggested that earlier... I'm pretty confident that its already pulled in through other terra components already.

gauthamsriman: So, Mike, would you prefer:

1) font-family: var(--terra-base-font-family, 'Roboto', sans-serif);

OR, just leave it hardcoded as before, like:

2) font-family: 'Roboto', sans-serif;

??

mhemesath: The ideal solution is that we dont set font-family here at all. Terra already does it in its base style we simply need set the custom property.

bjankord:
+1 on this change https://github.com/cerner/terra-consumer/pull/91/commits/219de0ffc72469aebcb87d6b458e8ccc43dd1e52
Looks like there is a merge conflict in that changelog that needs resolved.

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -15,7 +15,6 @@
 
   .layout {
     display: flex;
-    font-family: 'Roboto', sans-serif;
```
alex-bezek: this is non passive. you will break healthe at cerner doing this. can we set a default of roboto and allow memorial herman to override this? 
@mhemesath thoughts? should the library just not set this and require healthe@cerner to set this to roboto (as well as portal)

gauthamsriman: I am importing terra-base's styles in Layout.jsx as suggested by @bjankord and @mhemesath . Should I still keep this line?

mhemesath: @alex-bezek the point is to update healthe-at-cerner to configure the theme at the application level

alex-bezek: we discussed, this makes sense. just make sure you update the themes that were relying on this. I'll probably release this as a non passive change and bump the major version

PR #92, opened 2017-10-27
upgrade and fix pop up dependency

alex-bezek:
### Summary
Need to upgrade to use newer pop up and fix our usage of one of its internal files

### Additional Details
We had to hack together a quick fix at one point by consuming an internal file. This caused a breaking change even though they only had a minor version bump.

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/src/Nav.jsx
+++ b/packages/terra-consumer-nav/src/Nav.jsx
@@ -3,8 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Button from 'terra-button';
 import IconClose from 'terra-icon/lib/icon/IconClose';
-import TerraPopup from 'terra-popup';
-import PopupHeights from 'terra-popup/lib/PopupHeights';
+import Popup from 'terra-popup';
 import ResponsiveElement from 'terra-responsive-element';
 import NavItems from './components/nav-items/NavItems';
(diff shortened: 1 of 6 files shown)


packages/terra-consumer-nav/src/Nav.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames/bind';
 import Button from 'terra-button';
 import IconClose from 'terra-icon/lib/icon/IconClose';
 import TerraPopup from 'terra-popup';
-import PopupHeights from 'terra-popup/lib/_PopupHeights';
+import PopupHeights from 'terra-popup/lib/PopupHeights';
```
dkasper-was-taken: The new version of popup exposes this file in it's options, thus the name change from the private denoted '_'.
```
Popup.propTypes = propTypes;
Popup.defaultProps = defaultProps;
Popup.Opts = {
  attachmentPositions: ATTACHMENT_POSITIONS,
  heights: PopupHeights,
  widths: PopupWidths,
};

export default Popup;
```

alex-bezek: ahhh, that makes sense. Thanks for pointing this out!
I'll consume the variables that way instead of our hacky invasive non passive way.
we should be removing this anyways once we get the auto height + content that expands issue fix.

PR #98, opened 2017-11-02
Fixing burger bar without icon

kafkahw:
### Summary
This PR fixes the overlapping issue of nav burger icon when there's no mobile logo configured.

DETAIL: in the previous release, we set burger button's `position` to `absolute` which takes it off the flow. However, that will take the whole nav bar off if img is not the nav bar. So we're adding an img holder that has a fixed height to keep that position.  

@cerner/terra

### DIFF
--- a/packages/terra-consumer-layout/src/Layout.jsx
+++ b/packages/terra-consumer-layout/src/Layout.jsx
@@ -90,11 +90,11 @@ class Layout extends React.Component {
             <div className={cx('main-container-inner')}>
               <div className={cx('nav-burgerbar')}>
                 <Nav.Burger handleClick={this.toggleNav} />
-                {logo && logo.mobileLogo &&
-                  <div className={cx('mobile-logo')}>
+                <div className={cx('mobile-logo')}>
+                  {logo && logo.mobileLogo &&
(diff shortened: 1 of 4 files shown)


kafkahw:
Here's the issue:
issue

kafkahw:
here's the fix:
fixed

packages/terra-consumer-layout/src/Layout.jsx
```diff
@@ -90,11 +90,11 @@ class Layout extends React.Component {
             <div className={cx('main-container-inner')}>
               <div className={cx('nav-burgerbar')}>
                 <Nav.Burger handleClick={this.toggleNav} />
-                {logo && logo.mobileLogo &&
-                  <div className={cx('mobile-logo')}>
+                <div className={cx('mobile-logo')}>
```
kafkahw: The solution is that always display this img holder no matter if mobileLogo is existing or not. Assign a fixed height (35px which is the max-height of img) to this img holder to keep the flow.

mhemesath: NIT: isn't it more appropriate for the nav-burgerbar to have the height of 35px?

PR #99, opened 2017-11-03
Add storybook

zhongyn:
### Summary
Setup storybook as doc site, migrate nav/layout/icon examples to storybook
```
npm install
npm run storybook
visit localhost:8081
```


mhemesath:
I prefer that the stories live in the package of each component. We can then use [story discovery](https://storybook.js.org/basics/guide-react/#create-the-config-file) to pull all the stories into storybook.

mhemesath:
Can we update the the deployment to run storybook rather than site? I want to remove the site, but we can do that in a subsequent pull request when we switch everything to webdriver.io and storybook

bjankord:
@mhemesath Storybook just [added a feature](https://github.com/storybooks/storybook/issues/1013) to autoload all files with `/.story.js$/` in their name to their v3.3.0 release branch. v3.3.0 has not yet been released yet though, there are a few other issues in that [milestone](https://github.com/storybooks/storybook/milestone/9).

zhongyn:
@mhemesath putting the stories in the package of each component works for most components except those with translations. In this PR, all stories and story configs are inside the storybook folder and isolate from the outside.

- root
  - storybook
     - package.json
     - config.js
     - webpack.config.js
     - stories
  - packages
  - package.json

so it can install terra-consumer-nav and add terra-i18n-plugin in its own node_modules. terra-in8-plugin then aggregates the translations by searching the node_modules folder. i18n is required for component like [NavHelp](https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/nav-help/NavHelp.jsx#L38).

In the terra-badge example, the structure is:

- root
  - .storybook
    - config.js
  - packages
    - terra-badge
      - stories
  - package.json

That works. In the case of [terra-alert with translations](https://github.com/cerner/terra-core/tree/storybook/packages/terra-alert),  it requires adding webpack.config.js with terra-i18n-plugin inside .storybook, and it will try to search the root level node_modules for translations folder. That means terra-alert should be added as devDependency in the root level. Not sure if that's appropriate. Any idea for handling this case?

zhongyn:
yes, will update the deployment to run storybook after this PR

mhemesath:
> terra-i18n-plugin then aggregates the translations by searching the node_modules folder

That plugin is only needed for aggregation. Do we need aggregation with storybook? Why can't the component find the translations relative to its directory. If we do need i18n plugin, can the source folder be modified it uses to search for translations?

mhemesath:
I am really adamant about having the stories of the components live in each package if possible.

zhongyn:
> That plugin is only needed for aggregation. Do we need aggregation with storybook?

yes, terra-consumer-nav depends on terra-i18n, and i18n loads translation files from the aggregation folder. i18n-plugin generates the aggregation. Without the aggregation folder webpack compile will fail. One solution is creating the aggregation manually with empty message. If so terra-i18n will work but no message is pulled in the component.

> Why can't the component find the translations relative to its directory.

terra-i18n at the root level pulls in the translations. The component itself doesn't load the translations.

>  If we do need i18n plugin, can the source folder be modified it uses to search for translations?
 
i18n plugin only checks the node_modules for translation files[node_modules](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js#L47). We can update it to take a custom path param so that it can search any custom folder

zhongyn:
I created an issue in terra-core for extending i18n plugin with custom search path: https://github.com/cerner/terra-core/issues/998

package.json
```diff
@@ -87,19 +91,30 @@
...
     "selenium-server-standalone-jar": "^3.4.0",
     "shelljs": "^0.7.7",
+    "style-loader": "^0.19.0",
     "stylelint": "~8.0.0",
     "stylelint-config-sass-guidelines": "^3.0.0",
     "stylelint-scss": "1.5.1",
     "stylelint-suitcss": "^1.0.0",
+    "terra-base": "^2.6.1",
     "terra-i18n": "^1.4.0",
+    "terra-i18n-plugin": "file:../terra-core/packages/terra-i18n-plugin",
```
zhongyn: just for local dev, after this PR is merged we can update it with the new version

package.json
```diff
@@ -60,9 +60,12 @@
     "publish": "check-installed-dependencies && npm run compile && npm test && npm run props-table && lerna publish",
     "start": "cd packages/terra-consumer-site && npm run start",
     "start:express": "node scripts/express/app.js",
+    "storybook": "start-storybook -p 8081 -c .storybook",
     "test": "npm run jest && npm run nightwatch"
   },
   "devDependencies": {
+    "@storybook/addon-actions": "^3.2.14",
```
bjankord: You may want to update to `3.3.0-alpha.4`. Version `3.2` had issues with loading JSONP which is needed for requiring translations. This issue was fixed in v3.3.0 alphas. More info: https://github.com/storybooks/storybook/issues/1411

.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<script>
+  document.documentElement.setAttribute("dir", "ltr");
```
mhemesath: you can probably just do this in your main decorator
```js
addDecorator(story => <div dir="ltr"><Base locale="en-US">{story()}</Base></div>);
```
Terra might log a warning, but the attribute will still work even if its not on HTML tag

.storybook/webpack.config.js
```diff
@@ -0,0 +1,61 @@
+const path = require('path');
+const I18nAggregatorPlugin = require('terra-i18n-plugin');
+const i18nSupportedLocales = require('terra-i18n/lib/i18nSupportedLocales');
+const CustomProperties = require('postcss-custom-properties');
+const rtl = require('postcss-rtl');
+
+const customProperties = CustomProperties({ preserve: true, warnings: false });
+customProperties.setVariables({
```
mhemesath: is this duplicated from our other webpack?

zhongyn: it's copied from terra-consumer-site webpack config. Probably it's not necessary, will remove it and test the code

.storybook/preview-head.html
zhongyn: neat! will update

zhongyn: just found that it doesn't work for popup/modal rendered outside the `Base` tree. `NavHelp` looks like this after the change:
screen shot 2017-12-04 at 9 55 07 am
So we still need to put it in html tag

PR #101, opened 2017-11-21
Add custom callback to profile links for click events

jpruetting:
### Summary
Added new prop to profile links to allow for custom onClick callback functions.

### DIFF
--- a/packages/terra-consumer-nav/src/components/user-profile/ProfileLink.jsx
+++ b/packages/terra-consumer-nav/src/components/user-profile/ProfileLink.jsx
@@ -13,6 +13,7 @@ const propTypes = {
   * An optional function. When supplied, gets triggered on link click.
   */
   handleClick: PropTypes.func,
+  customOnClick: PropTypes.func,
   ...navElementShape,
 };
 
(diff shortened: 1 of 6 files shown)


packages/terra-consumer-nav/src/components/user-profile/ProfileLink.jsx
```diff
@@ -13,6 +13,7 @@ const propTypes = {
   * An optional function. When supplied, gets triggered on link click.
   */
   handleClick: PropTypes.func,
+  customOnClick: PropTypes.func,
```
mhemesath: why can't handleClick be used?

jpruetting: There's a stack of components that ends up explicitly setting handleClick to a function used to close the modal.

Nav renders UserProfile with onLinkClick = Nav.handleProfileLinkClick()
https://github.com/cerner/terra-consumer/blob/terra-consumer-nav%400.2.8/packages/terra-consumer-nav/src/Nav.jsx#L149

UserProfile renders ProfileLinks with handleClick = onLinkClick
https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/user-profile/UserProfile.jsx#L82

ProfileLinks renders ProfileLink with handleClick = handleClick
https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/user-profile/ProfileLinks.jsx#L80

ProfileLink renders SmartLink with handleClick = handleClick
https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/user-profile/ProfileLink.jsx#L33

SmartLink renders NavLink with onClick = handleClick
https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/smart-link/SmartLink.jsx#L45

jpruetting: As far as I can tell, the chain never opens up to allow anything other than that one function to be used.

alex-bezek: While this is gross, I think it makes the public API pretty close to our ideal state, and we just need to figure out a clean way to handle the internal state. We keep tacking on requirements without taking the time to refactor appropriately so this is getting out of control. At some point in the future, we probably need to sit down and determine all our existing requirements, any forseeable requirements, and refactor pretty heavily

PR #103, opened 2017-12-07
Updated help button content element to allow user copying content in subitem

Mahmud-Khan:
### Summary
Support/request ID not copyable after UI Uplift. Each item in Help pop is a button. When there is subitem clicking the button toggle the content. Since the toggle content is inside a button user cannot select the text inside the button. There is requirement for Portal to allow the user to copy a support id (ex. 1be787bfa7594928d4f5de2991c399ab) from the popup. 

Updated the Help popup/modal content component by separating the toggle header and toggle content. Toggle content is a paragraph, so the user can copy the content.Also added overflow-wrap CSS  property for the content, because the long string like this id 1be787bfa7594928d4f5de2991c399ab is not wrapping.

**Note**: 
**overflow-wrap: break-word**  is working for chrome,safari & firefox but not in ie/edge. Please let me know if you have any suggestion to wrap the long string like support ID mentioned above.

**Before**:

helpbuttontoggle_before

**After**

helpbuttontoggle_after

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,4 +1,11 @@
 ChangeLog
+# 0.2.10 - (December 07, 2018)
+### Changed
+- Updated the help button content with toggle.Toggle content is separated from
+- toggle header.
+
+
+------------------
 # 0.2.9 - (November 21, 2017)
 ### Added
 - Added custom callback functions to profile links for click events
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,9 +1,8 @@
 ChangeLog
-# 0.2.10 - (December 07, 2018)
+# 0.2.10 - (December 07, 2017)
 ### Changed
 - Updated the help button content with toggle.Toggle content is separated from
 - toggle header.
-
 
 ------------------
 # 0.2.9 - (November 21, 2017)
--- a/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
+++ b/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
@@ -47,8 +47,10 @@
     }
   }
 
-  .help-item-border {
-    border-top: 1px solid var(--terra-consumer--nav-help-item-border-color, #dedfe0);
+  .border-top-except-first-element {
+    &:not(:first-child) {
+      border-top: 1px solid var(--terra-consumer--nav-help-item-border-color, #dedfe0);
+    }
   }
 
   .help-item-text {
--- a/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
+++ b/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
@@ -37,6 +37,10 @@
     text-decoration: none;
     width: 100%;
 
+    &:not(:first-child) {
+      border-top: 1px solid var(--terra-consumer--nav-help-item-border-color, #dedfe0);
+    }
+
     &:hover {
       background: var(--terra-consumer--nav-help-item-hover-background, #eee);
       cursor: pointer;
@@ -47,7 +51,7 @@
     }
   }
 
-  .border-top-except-first-element {
+  .help-item-wrapper {
     &:not(:first-child) {
       border-top: 1px solid var(--terra-consumer--nav-help-item-border-color, #dedfe0);
     }
--- a/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
+++ b/packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
@@ -30,7 +30,6 @@
   .help-item {
     background: var(--terra-consumer--nav-help-item-background, #fff);
     border: 0;
-    border-bottom: 1px solid var(--terra-consumer--nav-help-item-border-color, #dedfe0);
     border-radius: 0;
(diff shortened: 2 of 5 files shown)


packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 ChangeLog
+# 0.2.10 - (December 07, 2018)
```
KevinJJackson: it's not 2018 yet, almost!

packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
...
+### Changed
+- Updated the help button content with toggle.Toggle content is separated from
+- toggle header.
+
+
```
KevinJJackson: np: extra space

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -46,10 +46,10 @@ class NavHelpContent extends React.Component {
       const toggleIcon = isOpen ? <IconChevronUp /> : <IconChevronDown />;
 
       if (content.children && content.children.length > 0) {
-        contentElement = (<Button
-          key={`${content.text}`}
+        const helpItemStyle = cx('help-item', index > 0 ? 'help-item-border' : '');
```
KevinJJackson: change class def to `'help-item', index > 0 && 'help-item-border'`

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -46,10 +46,10 @@ class NavHelpContent extends React.Component {
...
+        contentElement = (<div key={`${content.text}`} ><Button
```
KevinJJackson: curious, does content.text need to inside a template string?

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -58,6 +58,7 @@ class NavHelpContent extends React.Component {
             fill={<div><SafeHtml text={content.text} /></div>}
             fitEnd={<div>{toggleIcon}</div>}
           />
+        </Button>
           <Toggler isOpen={isOpen} isAnimated={false} className={cx('toggler-padding')}>
             { content.children.map(element => (
               <p key={`${element.text}`} className={cx('toggler-content-alignment')}>
```
KevinJJackson: same template string comment

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -67,11 +68,12 @@ class NavHelpContent extends React.Component {
               </p>))
           }
           </Toggler>
-        </Button>);
+        </div>);
       } else {
+        const helpItemStyle = cx('help-item', index > 0 ? 'help-item-border' : '');
```
KevinJJackson: same class def comment

KevinJJackson: maybe move the style def outside the if/else if it's always going to be applied in both scenarios

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
```diff
@@ -44,12 +44,12 @@ class NavHelpContent extends React.Component {
       let contentElement;
       const isOpen = this.state.togglers[index];
       const toggleIcon = isOpen ? <IconChevronUp /> : <IconChevronDown />;
+      const helpItemStyle = cx('help-item', index > 0 && 'help-item-border');
```
mhemesath: can we not apply the border rule with CSS rather than logically adding another class name?

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -48,6 +47,10 @@
     }
   }
 
+  .help-item-border {
```
mhemesath: class names should describe what a thing is, not what it looks like. From the code it looks like this is for n+1 help item... which I'd ask if we can use a pseudo selector to do that instead of an additional class name

packages/terra-consumer-nav/src/components/nav-help/NavHelpContent.jsx
mhemesath: Also the correct way to do conditional class names is
```js
cx({ 'help-item-border': index  > 0 })
```
https://www.npmjs.com/package/classnames#usage

KevinJJackson:
@Mahmud-Khan you may be able to use the word-break css property for ie https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -48,6 +47,12 @@
     }
   }
 
+  .border-top-except-first-element {
```
mhemesath: Do not name classes after what they are styling, name them after what they are doing.  Describe the container element this is applied to and style that.

mhemesath: Ah I see this is reused across components... don't use CSS classes like that. Put this style onto each component that needs it.

PR #104, opened 2017-12-18
Updated help button and help-item hover background

Mahmud-Khan:
### Summary
- Updated the help button and help-item hover background to make it consistent with profile item popup.
- Updated the margin and padding for toggle content in help popup based on UX feedback.

Before:

help_button_hover_before

After:

help_button_hover_after

@cerner/terra

### DIFF
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,4 +1,11 @@
 ChangeLog
+# 0.2.11 - (December 18, 2017)
+### Changed
+- Updated the hover background for help-button and popup content to make it
+- consistent with profile button and content.
+- Updated the margin and padding for toggle content in help popup.
+
+------------------
 # 0.2.10 - (December 07, 2017)
 ### Changed
 - Updated the help button content with toggle.Toggle content is separated from
--- a/packages/terra-consumer-nav/CHANGELOG.md
+++ b/packages/terra-consumer-nav/CHANGELOG.md
@@ -1,8 +1,7 @@
 ChangeLog
-# 0.2.11 - (December 18, 2017)
+# 0.2.11 - (December 21, 2017)
 ### Changed
-- Updated the hover background for help-button and popup content to make it
-- consistent with profile button and content.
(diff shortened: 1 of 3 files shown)


packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
```diff
@@ -85,7 +91,8 @@
   }
 
   .toggler-content-alignment {
-    padding: 0 40px 20px 20px;
+    margin: 0;
+    padding: 10px 40px 20px 20px;
```
KevinJJackson: Ux really wants 10px above and 20px below? Seems strange, might want to have them look at this before this gets merged in.

Mahmud-Khan: If we don't add this extra 10px it does not look good when the button is clicked in chrome. The selection border in chrome is very close to the content.Please check the  screenshot below

screen shot 2017-12-19 at 4 12 47 pm

Got +1 from Anna and Julie in the Jira.

KevinJJackson: i was more worried about the 20 below not the 10 above

packages/terra-consumer-nav/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 ChangeLog
+# 0.2.11 - (December 18, 2017)
+### Changed
+- Updated the hover background for help-button and popup content to make it
+- consistent with profile button and content.
```
KevinJJackson: don't split updates, makes the list a little strange to read

packages/terra-consumer-nav/src/components/nav-help/NavHelp.scss
Mahmud-Khan: If we consider the whole help-item including toggle button and content there is 20px both top & bottom when the toggle is open. 
screen shot 2017-12-20 at 9 40 09 am

packages/terra-consumer-nav/CHANGELOG.md
zhongyn: +1

PR #107, opened 2018-01-04
remove missing logo icon on certain browsers

KevinJJackson:
### Summary
Change nav logo to render a p tag if no url is passed to avoid missing image icon in chrome/safari/ie

Before:

image

After:

image

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
@@ -43,10 +43,17 @@ const NavLogo = ({
 }) => {
   const imgProps = {
     className: cx('img'),
-    ...url && { src: url },
+    ...url ? {
+      src: url,
+      alt: altText,
+    } : {
+      children: altText,
+    },
   };
 
-  const imgTag = (<img alt={altText} {...imgProps} />);
+  const LogoElement = url ? 'img' : 'p';
+
+  const imgTag = (<LogoElement {...imgProps} />);
   const imgContent = (link) ? <SmartLink {...link}>{imgTag} </SmartLink> : imgTag;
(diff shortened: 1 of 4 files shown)


packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -43,10 +43,17 @@ const NavLogo = ({
...
+    ...url ? {
+      src: url,
+      alt: altText,
+    } : {
+      children: altText,
+    },
   };
 
-  const imgTag = (<img alt={altText} {...imgProps} />);
+  const LogoElement = url ? 'img' : 'p';
```
alex-bezek: instead of a <p> should it just be null?

KevinJJackson: imgTag has to be something since that's what's being displayed in the smartlink, figured a 'p' tag made the most sense for text

packages/terra-consumer-nav/src/components/nav-logo/NavLogo.jsx
```diff
@@ -43,10 +43,17 @@ const NavLogo = ({
 }) => {
   const imgProps = {
     className: cx('img'),
-    ...url && { src: url },
+    ...url ? {
+      src: url,
+      alt: altText,
+    } : {
+      children: altText,
```
zhongyn: where is this children rendered?

zhongyn: i see, it is rendered inside p tag

KevinJJackson: this is rendered inside the 'p' tag generated inside `LogoElement` when a url is not passed

PR #109, opened 2018-01-18
refreshing sub navigation opens up tree navigation

divyaashritha:
### Summary
Open the navigation tree when redirected to one of its subnav from dashboard

JIRA: [PORTALDEV-41083](https://jira2.cerner.com/browse/PORTALDEV-41083)

### Testing
Tested locally by changing terra-consumer-nav to point to this branch in [package.json](https://github.cerner.com/HealtheLife/healthelastic-react/blob/PORTALDEV-41083/package.json#L43) and also passed the currentURL prop from [healthelastic-react](https://github.cerner.com/HealtheLife/healthelastic-react/commit/96b3f4645abf3d9d1fffdcc09c53c03eff9ee3e8)

### DIFF
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
@@ -57,7 +57,7 @@ class NavItems extends Component {
     /* if there is a sub nav which matches the currentURL open the parent nav */
     currentComponent.props.navItems.forEach((element, i) => {
       if (element.subItems) {
-        /* Use polyfill to support find() in IE*/
+        /* Use polyfill to support find() in IE */
         const match = element.subItems.find(item => item.url === currentURL);
         if (match) {
           currentComponent.setState({ openToggle: i });
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
@@ -47,23 +47,25 @@ class NavItems extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props !== nextProps) {
-      const currentComponent = this;
-      const currentURL = nextProps.currentURL;
-
-      /* if there is a sub nav which matches the currentURL open the parent nav */
-      currentComponent.props.navItems.forEach((element, i) => {
-        if (element.subItems) {
-          element.subItems.forEach((item) => {
-            if (currentURL === item.url) {
-              currentComponent.setState({
-                openToggle: i,
-              });
-            }
-          });
-        }
-      });
+    if (this.props.currentURL === nextProps.currentURL) {
+      return;
     }
+
+    const currentComponent = this;
+    const currentURL = nextProps.currentURL;
+
+    /* if there is a sub nav which matches the currentURL open the parent nav */
+    currentComponent.props.navItems.find((element, i) => {
+      if (element.subItems) {
+        element.subItems.find((item) => {
+          if (currentURL === item.url) {
+            currentComponent.setState({
+              openToggle: i,
+            });
+          }
+        });
+      }
+    });
   }
 
   handleToggle(toggleId, isOpen) {
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
@@ -55,15 +55,13 @@ class NavItems extends Component {
     const currentURL = nextProps.currentURL;
 
     /* if there is a sub nav which matches the currentURL open the parent nav */
-    currentComponent.props.navItems.find((element, i) => {
+    currentComponent.props.navItems.forEach((element, i) => {
       if (element.subItems) {
-        element.subItems.find((item) => {
-          if (currentURL === item.url) {
-            currentComponent.setState({
-              openToggle: i,
-            });
-          }
-        });
+        /* Use polyfill to support find() in IE*/
+        const match = element.subItems.find(item => item.url === currentURL);
+        if (match) {
+          currentComponent.setState({ openToggle: i });
+        }
       }
     });
   }
--- a/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
+++ b/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
@@ -51,7 +51,8 @@ class NavItems extends Component {
       const currentComponent = this;
       const currentURL = nextProps.currentURL;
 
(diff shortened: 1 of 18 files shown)


packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
```diff
@@ -42,14 +46,34 @@ class NavItems extends Component {
...
+      const currentComponent = this;
+      const currentURL = nextProps.currentURL;
+
+      /* if there is a sub nav which matches the currentURL open the parent nav */
+      currentComponent.props.navItems.forEach((element, i) => {
+        if (element.subItems) {
+          element.subItems.forEach((item) => {
+            if (currentURL === item.url) {
+              currentComponent.setState({
+                openToggle: i,
```
alex-bezek: There should be only 1 match right? Does it need to keep looping through the rest of them?

alex-bezek:
Is this not already do-able using the isActive attribute?
https://github.com/cerner/terra-consumer/blob/526f4d4873cc90b5e9d66a3f15470a7cd2007c2a/packages/terra-consumer-nav/src/components/nav-items/NavItem.jsx#L46

The patient portal test example here already works
https://terra-consumer.herokuapp.com/static/#/tests/layout-tests/portal
you can see the config here
https://github.com/cerner/terra-consumer/blob/526f4d4873cc90b5e9d66a3f15470a7cd2007c2a/packages/terra-consumer-layout/tests/nightwatch/PortalLayout.jsx#L117

And patient portal is using this, and I thought this worked for them. What am i missing? Is it some react router difference?

packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
```diff
@@ -42,14 +46,34 @@ class NavItems extends Component {
     this.handleToggle = this.handleToggle.bind(this);
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props !== nextProps) {
```
kafkahw: this is a shallow comparison, which is not good. Should compare attribute directly. Also in order to save indentations should use return statement first.
{code}
if (this.props.currentURL === nextProps.currentURL)
    return;
// The rest of code starts here.
{code}

divyaashritha: Updated!

packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
divyaashritha: Good catch! I have updated to find instead of forEach

zhongyn:
@alex-bezek the usage of `isActive` allows the navitem to be toggled [when mounted](https://github.com/cerner/terra-consumer/blob/master/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx#L36-L39). I guess in dex the scenario is toggling navitems after mounted.
 So there are two possible solutions for this:
1. reuse `isActive` prop, dex will generate this key for every nav item in its data contract at runtime and pass the data to terra-consumer-layout. Also NavItems would need similar logic in `componentWillReceiveProps` like this PR
2. solution proposed in this PR

Both solutions require the logic of checking the open state of subnav in NavItems component. Solution 2 would require less work for dex to update their data contract.

packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx
```diff
@@ -42,14 +46,36 @@ class NavItems extends Component {
...
+      return;
+    }
+
+    const currentComponent = this;
+    const currentURL = nextProps.currentURL;
+
+    /* if there is a sub nav which matches the currentURL open the parent nav */
+    currentComponent.props.navItems.find((element, i) => {
+      if (element.subItems) {
+        element.subItems.find((item) => {
```
kafkahw: this is not the right way to use `find`. the callback passed to `find` should return a boolean value. Code should be changed to something like the following:
```
const match = elements.subItems.find(item=> item.url === currentURL);
if (match) {
  currentComponent.setState({openToggle: i});
}
```

kafkahw: Also, since `find` method is not natively supported in IE (see [ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Browser_compatibility)), you may need add some polyfill for it and should test it in IEs that terra-consumer supports.

divyaashritha: Updated!  Also, as our application is consuming this package and uses polyfill it supports {code}find{code} method in IE, but if someone wants to consume this polyfill is required.

alex-bezek:
https://github.com/cerner/terra-consumer/pull/109#discussion_r162710389
https://github.com/cerner/terra-consumer/pull/109#issuecomment-359586313
concerning these comments. 
We are saying there is already a prop called `isActive` but updating it doesn’t have any affect because its only used to calculate `initialIndex` in the constructor (so on mount) because at the time, only portal needed it, and it reloaded the page. 

So to get something that changes each time the prop is updated, we are opting to add a new prop instead of making the existing prop update on componentwillreceiveprops or something? This almost seems like you should just fix the bug of "Prop doesn't acurrately update components when changed" instead of adding a new prop.


> dex will generate this key for every nav item in its data contract at runtime and pass the data to terra-consumer-layout. 

Dex already uses react routers `match.url` to determine isActive the first time. And its propably updating as you bounce aroud anyways (this layout just isn't re-calculating based off it). 

> Also NavItems would need similar logic in componentWillReceiveProps like this PR
isn't this handled in healthelastic-react when it loops over all the navigation items, compares their url to the react routers url, and sets isActive based off that?

zhongyn:
> Dex already uses react routers match.url to determine isActive the first time. And its propably updating as you bounce aroud anyways (this layout just isn't re-calculating based off it).

If dex already gererates `isActive` in its data contract. Then it makes sense to reuse this prop. Initially I thought only portal utilizes the prop. 

> isn't this handled in healthelastic-react when it loops over all the navigation items, compares their url to the react routers url, and sets isActive based off that?

Even `isActive` is set in healthelastis-react, it still requires toggling the navItems in `componentWillReceiveProps` based on the value of `isActive`. It's one of the two parts:
- determine the value of `isActive`
- toggle the navitem based on `isActive`

In sum, the logic of determining the value of `isActive` can be bubbled up to healthelastic and then the logic of toggling the navItem will be inside `componentWillReceiveProps` similar to [this line](https://github.com/divyaashritha/terra-consumer/blob/a0715aa12d8bf2208fe0cdfadea7b371aa6d73d1/packages/terra-consumer-nav/src/components/nav-items/NavItems.jsx#L40)

PR #110, opened 2018-01-21
Remove roboto and update examples

mhemesath:
### Summary
Remove roboto from being hardcoded imported. This forces all consumers to incur the cost of loading the roboto font even if its not being used. Roboto should be configured by the consuming application to be loaded dynamically using https://github.com/typekit/webfontloader


zhongyn:
changelog needs update inorder to pass travis build

PR #121, opened 2019-01-29
Update terra deps to latest major release

bjankord:
### Summary
* Update terra dependencies to latest stable release
  * (except terra-toolkit) we should keep this as-is as it's only a dev-depdency
* Update zIndex on terra-consumer-layout overlay


bjankord:
Closing in favor of a smaller update

PR #122, opened 2019-01-29
Fix nav overlay zindex issue

bjankord:
### Summary
Resolves z-index issue with terra-overlay and terra-consumer-nav

Before screenshot
screen shot 2019-01-29 at 5 00 34 pm

After screenshot
screen shot 2019-01-29 at 5 00 45 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     ],
     "moduleNameMapper": {
       "\\.(css|scss)$": "identity-obj-proxy"
-    }
+    },
+    "testURL": "http://localhost"
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -34,7 +34,8 @@
     ],
     "moduleNameMapper": {
       "\\.(css|scss)$": "identity-obj-proxy"
-    }
+    },
+    "testURL": "http://localhost"
```
bjankord: This is needed due to a JSDom update that jest pulls in.
More info here: https://github.com/jsdom/jsdom/issues/2304

packages/terra-consumer-layout/src/Layout.scss
```diff
@@ -27,7 +27,7 @@
     min-width: var(--terra-consumer-nav-width, 320px);
     position: relative;
     width: var(--terra-consumer-nav-width, 320px);
-    z-index: 200;
+    z-index: 99;
```
bjankord: This allows the nav to site behind the overlay

PR #124, opened 2019-02-07
Update DOM order in layout

bjankord:


roxjcalderon:
We're seeing issues with the utility/settings menu overlay in Chrome and Edge.

